PR #4, opened 2020-10-05
[TASK-3] [Feature] As a user, I can sign-in into the app using my email and password

minhnimble:
https://github.com/minhnimble/android-surveys-list/projects/1#card-46080084

## What happened

When signing-in, users will be redirected to the empty home dashboard.
 
## Insight

**Acceptance Criteria**
- [x] Connect to login API

**Testing and implementation resources**
- The following credentials can be used: minh@nimblehq.co - 12345678
 
## Proof Of Work
Screen Recording 2020-10-01 at 17 30 18

PR #7, opened 2020-10-06
[TASK-3] [Chore] As a user, I can sign-in into the app using my email and password

minhnimble:
https://github.com/minhnimble/android-surveys-list/projects/1#card-46080084

## What happened

Fixing multiple comments from the previous logic branch #4:
@sleepylee :
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499516125

@hoangnguyen92dn : 
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499480318
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499483011
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499491347
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499492165
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499492252
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499493678
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499502920
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499505347
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499507671
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499509029
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499522109
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499526813
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499528001

@manh-t : 
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499629236
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499638777
- https://github.com/minhnimble/android-surveys-list/pull/4#discussion_r499644597

Create this new refactor branch for help isolate the comments fixes from the logic branch and make the review process easier 💪
 
## Insight

**Acceptance Criteria**
- [x] All comments are checked and fixed.
 
## Proof Of Work

No visible changes, just under the hood refactoring 😁.


app/src/test/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModelTest.kt
```diff
@@ -101,34 +95,27 @@ class SignInViewModelTest {
...
 
         // Act
-        signingViewModel
-            .login("test@nimblehq.co", "12345678")
-            .test()
+        signingViewModel.inputs.email("test@nimblehq.co")
+        signingViewModel.inputs.password("12345678")
+        signingViewModel.login()
 
         val showLoadingObserver = signingViewModel.showLoading.test()
```
sleepylee, line 118: sometimes I'd prefer to group these attestations into the same test case where we are manipulating the same source, for example: the _loading is being emitted via doOnSubscribe and doFinally of an api request, then it makes sense to assert the state within the same test case where we are testing the behavior of that api invoker observable source too

sleepylee (commented):
rest lgtm

app/src/main/java/co/nimblehq/usecase/session/LoginByPasswordSingleUseCase.kt
```diff
@@ -21,7 +22,7 @@ class LoginByPasswordCompletableUseCase @Inject constructor(
         val password: String
     )
 
-    override fun create(input: Input): Completable {
+    override fun create(input: Input): Single<OAuthResponse> {
```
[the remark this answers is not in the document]
minhnimble: oh I understand your point, totally agree. let me work on it

app/src/test/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModelTest.kt
```diff
@@ -13,34 +18,37 @@ import org.junit.Test
...
         whenever(
-            mockLoginByPasswordCompletableUseCase.execute(any())
-        ) doReturn Single.error<LoginError>(LoginError(null)).ignoreElement()
+            mockLoginByPasswordSingleUseCase.execute(any())
+        ) doReturn Single.error(LoginError(null))
 
         // Act
-        val testSubscriber = signingViewModel
-            .login("invalid@nimblehq.co", "12345678")
+        val isLoginSuccessObserver = signingViewModel
```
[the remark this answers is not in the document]
minhnimble: hmm, it is actually the flag name, but yeah the test case is negative test, let me update it

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInFragment.kt
```diff
@@ -60,43 +60,44 @@ class SignInFragment: BaseFragment(), BaseFragmentCallbacks {
...
     override fun bindViewEvents() {
         btSignInLogin.subscribeOnClick {
-            viewModel.login(
-                email = etSignInEmail.text.toString(),
-                password = etSignInPassword.text.toString()
-            ).subscribeBy(
-                onComplete = :: openMainActivity,
-                onError = ::handleError
-            ).bindForDisposable()
+            viewModel.login()
```
hoangnguyen92dn, line 63: Not a big deal but using function reference looks better

```
        btSignInLogin
            .subscribeOnClick(viewModel::login)
            .bindForDisposable()
```

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInFragment.kt
```diff
@@ -60,43 +60,44 @@ class SignInFragment: BaseFragment(), BaseFragmentCallbacks {
...
             .bindForDisposable()
 
         viewModel.showLoading
             .subscribe(::bindLoading)
             .bindForDisposable()
     }
 
-    private fun bindLoading(isLoading: Boolean) {
-        toggleLoading(isLoading)
+    private fun bindEnableLoginBtn(isEnabled: Boolean) {
```
hoangnguyen92dn, line 89: Imo, extract enable/disable to method seems unnecessary 😅   We create method for `bindLoading` because we could do more login while loading in progress (show/hide views, or other stuff)

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInFragment.kt
```diff
@@ -60,43 +60,44 @@ class SignInFragment: BaseFragment(), BaseFragmentCallbacks {
...
+    private fun bindLoading(isLoading: Boolean) {
+        toggleLoading(isLoading)
     }
 
-    private fun openMainActivity() {
-        navigator.navigateToMainActivity()
+    private fun bindLoginStatus(error: Throwable) {
+        when (error) {
+            is LoginError -> { displayError(error) }
+            else -> { navigator.navigateToMainActivity() }
```
hoangnguyen92dn, line 100: We could remove braces here 😅

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
```diff
@@ -40,28 +47,36 @@ class SignInViewModel @ViewModelInject constructor(
...
 
-    fun login(
-        email: String,
-        password: String
-    ): Completable {
-        return loginByPasswordCompletableUseCase
-            .execute(LoginByPasswordCompletableUseCase.Input(email, password))
+    fun login() {
+        loginByPasswordSingleUseCase
+            .execute(LoginByPasswordSingleUseCase.Input(_email.value ?: "", _password.value ?: ""))
```
hoangnguyen92dn, line 58: Please use extension `orEmpty()` to get rid of `?: ""` 👨‍🚀   `_email.value.orEmpty()`

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
```diff
@@ -40,28 +47,36 @@ class SignInViewModel @ViewModelInject constructor(
...
             .doOnSubscribe { _showLoading.onNext(true) }
+            .flatMapCompletable { response ->
+                updateTokenCompletableUseCase.execute(UpdateTokenCompletableUseCase.Input(response))
+            }
```
hoangnguyen92dn, line 62: Optimize a little bit by using `map` from rx

```
            .doOnSubscribe { _showLoading.onNext(true) }
            .map(UpdateTokenCompletableUseCase::Input)
            .flatMapCompletable(updateTokenCompletableUseCase::execute)
```

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
```diff
@@ -40,28 +47,36 @@ class SignInViewModel @ViewModelInject constructor(
...
             .doFinally { _showLoading.onNext(false) }
+            .subscribe(
+                { _isLoginSuccess.onNext(Ignored(null)) },
+                { _isLoginSuccess.onNext(it) }
```
hoangnguyen92dn, line 66: Why do we need to emit `Ignored(null)` here? `_isLoginSuccess` but we are emitting error to this Subject seems not correct. If you wanna handle error, we could use `subscribeBy`

```
            .doFinally { _showLoading.onNext(false) }
            .subscribeBy(
                onError = // Handle error here
            )
            .bindForDisposable()
```

app/src/test/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModelTest.kt
minhnimble: Personally, I think we should separate the test cases for loginStatus and showLoading, even though it is called in the same observable flow. Because regardless of the use cases run successfully or failed, the loading state flow is the same, which is hiding the loading at the end. If we incorporate it into the same test case of the api request, it will act like chaining logics and I would rather clone the setup code than have chaining logic in one test function. A unit test should be testing one simple case only. What do you think? 🤔

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
```diff
@@ -40,28 +47,36 @@ class SignInViewModel @ViewModelInject constructor(
...
-    override fun updateInitialized(value: Boolean) {
-        _firstInitialized = value
+    override fun password(input: String) {
+        _password.onNext(input)
     }
 
-    override fun updatePassword(password: String) {
-        _password.onNext(password)
+    override fun initialized(input: Boolean) {
+        _firstInitialized = input
```
hoangnguyen92dn, line 80: Hmmm. I'd suggest to make `_firstInitialized` to public and access directly.

data/src/main/java/co/nimblehq/data/service/response/OAuthResponse.kt
```diff
@@ -1,24 +1,21 @@
 package co.nimblehq.data.service.response
 
-import android.os.Parcelable
 import com.google.gson.annotations.SerializedName
-import kotlinx.android.parcel.Parcelize
 
-@Parcelize
-data class OAuthResponse(val data: OAuthDataResponse): Parcelable
+data class OAuthResponse(
+	@SerializedName("data")val data: OAuthDataResponse
```
hoangnguyen92dn, line 6: Space here is 1 tab but below is 8 spaces? 🤔  And we need space between `("data")` and `val data`

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInFragment.kt
minhnimble: thanks man, I am not aware of this syntax, let me update 👍🙏

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
minhnimble: oh got it, let me update

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
minhnimble: we need `Ignored(null)` for the success case to trigger, to navigate to MainActivity. Because when using a method like showLoading, have to handle both error and success case, and in the error case we need to return a Throwable for message displaying.

Yeah I can definitely update to `subscribeBy`

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
minhnimble: I am following our conventions that we have a private var and public accessor accordingly, not sure why you suggest this...

data/src/main/java/co/nimblehq/data/service/response/OAuthResponse.kt
minhnimble: oh I guess it is my AS indentation again, I thought that I update it to 4 already... let me check again, thanks for spotting!

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
minhnimble: nice, thanks I will update

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInFragment.kt
minhnimble: oh I thought we need, let me update, thank you!

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInFragment.kt
minhnimble: hmm, this is for preparing for future usage and consistency, I think it is fine to leave it here 🤔

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
hoangnguyen92dn: For our current project, we created NavigationEvent for trigger navigation to next screen if needed. So in your case, I'd suggest to implement like this

```
            .subscribeBy(
                onComplete = _navigationSubject.onNext(NavigationEvent.Main),
                onError = _errorSubject::onNext
            )
```

I think you could refactor to handle navigation later.

minhnimble: oh thanks for clearing this out to me, this makes sense to me now. Will definitely update it!

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
hoangnguyen92dn: As I can see, we created private Subject for emitting event from ViewModel and public Observable for observing new event from Fragment. So that why's we convert from Subject to Observable by using `get() = `, but `_firstInitialized` is primitive data type so we could use it directly, imo.

minhnimble: hmm, @sleepylee what is your opinion on this, it is indeed easier to access the primitive value and reduce the code base quite a lot, but it is not consistent with other part of the app.

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInFragment.kt
```diff
@@ -59,44 +59,45 @@ class SignInFragment: BaseFragment(), BaseFragmentCallbacks {
...
+        toggleLoading(isLoading)
     }
 
-    private fun openMainActivity() {
-        navigator.navigateToMainActivity()
+    private fun bindLoginStatus(error: Throwable) {
+        when (error) {
+            is LoginError -> displayError(error)
+            else -> navigator.navigateToMainActivity()
+        }
```
manh-t, line 101: This looks confusing imo, it looks like if we get other error type, app will navigate to Main Activity 😂 . Should we create an `error` variable to handle the error here? 🙏

app/src/main/java/co/nimblehq/usecase/session/UpdateTokenCompletableUseCase.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    override fun create(input: Input): Completable {
+        return Completable.fromAction {
+            secureStorage.apply {
+                userAccessToken = input.response.accessToken
+                userAccessTokenCreatedAt = input.response.createdAt
+                userAccessTokenExpiresIn = input.response.expiresIn
+                userRefreshToken = input.response.refreshToken
+                userTokenType = input.response.tokenType
+            }
```
manh-t, line 33: optional: we could use `with(input.response)` here 🙆‍♂️

minhnimble: oh I think it is good to update, let me refactor it 👍

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInFragment.kt
minhnimble: oh we have a discussing with @hoangnguyen92dn about this, this check will be clear when we update to use navigator, thanks for mentioning 👍

manh-t: yah, please put a TODO comment if we need to do something later 😸  it also helps me know that the issue has raised so we do not make the duplicate. Thank you 🙇

app/src/test/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModelTest.kt
sleepylee: well, I understood that for many cases we should separate the tests cases for sure, but like I said, due to our current implementation to manipulate the `_loading` stream as a side-effect thing, then it's ok to add it to the test case of the main flow too, a.k.a login() 🙏  but if you want to do this it's ok as well, as you can see, it's just that the test setup will have to be duplicated and the number of test cases as well,

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
sleepylee: https://kotlinlang.org/docs/reference/coding-conventions.html#property-names only use `_privateField` if you are generating a backing field :D so if we turn it into private only -> no need `_underscore`, but if we introduce `get()` from the backing field, then yes.

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInFragment.kt
minhnimble: got it, will add a TODO 👍

app/src/main/java/co/nimblehq/usecase/session/ClearTokenCompletableUseCase.kt
```diff
@@ -0,0 +1,27 @@
...
+import co.nimblehq.usecase.base.CompletableUseCase
+import io.reactivex.Completable
+import javax.inject.Inject
+
+class ClearTokenCompletableUseCase @Inject constructor(
+    rxSchedulerProvider: RxSchedulerProvider,
+    private val secureStorage: SecureStorage
+) : CompletableUseCase<Unit>(
+    rxSchedulerProvider.io(),
+    rxSchedulerProvider.main(),
```
manh-t, line 15: this use case doesn't interact with UI, so I'd suggest to use `rxSchedulerProvider.io()` as well 🙏

app/src/test/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModelTest.kt
minhnimble: okay, this case I think it is okay to combine, let me update

app/src/main/java/co/nimblehq/usecase/session/UpdateTokenCompletableUseCase.kt
```diff
@@ -0,0 +1,38 @@
...
+import co.nimblehq.usecase.base.CompletableUseCase
+import io.reactivex.Completable
+import javax.inject.Inject
+
+class UpdateTokenCompletableUseCase @Inject constructor(
+    rxSchedulerProvider: RxSchedulerProvider,
+    private val secureStorage: SecureStorage
+) : CompletableUseCase<UpdateTokenCompletableUseCase.Input>(
+    rxSchedulerProvider.io(),
+    rxSchedulerProvider.main(),
```
manh-t, line 17: we can modify here as well, I think

manh-t (commented):
rest LGTM 💯

app/src/main/java/co/nimblehq/usecase/session/UpdateTokenCompletableUseCase.kt
minhnimble: agree 👍

app/src/main/java/co/nimblehq/usecase/session/ClearTokenCompletableUseCase.kt
minhnimble: oh nice catch there, let me update

manh-t (approved):

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
```diff
@@ -1,29 +1,37 @@
...
 
-    private val _email = BehaviorSubject.createDefault("")
+    private val _email = BehaviorSubject.create<String>()
+
+    private val _password = BehaviorSubject.create<String>()
 
     private var _firstInitialized = true
 
-    private val _password = BehaviorSubject.createDefault("")
+    private val _loginStatus = PublishSubject.create<Throwable>()
```
hoangnguyen92dn, line 34: I'd suggest to rename this to `_error` for common usage

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
```diff
@@ -1,29 +1,37 @@
...
+    private val loginByPasswordSingleUseCase: LoginByPasswordSingleUseCase,
+    private val updateTokenCompletableUseCase: UpdateTokenCompletableUseCase
 ) : BaseViewModel(), Inputs {
 
-    private val _email = BehaviorSubject.createDefault("")
+    private val _email = BehaviorSubject.create<String>()
+
+    private val _password = BehaviorSubject.create<String>()
 
     private var _firstInitialized = true
```
hoangnguyen92dn, line 32: Please move this to bottom of variables, we could define `val` and `Subject` to top of variable scope.

hoangnguyen92dn (approved):
LGTM

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
minhnimble: yep, this is actually a throwable, so error would make sense. I will update in the next PR 👍

app/src/main/java/co/nimblehq/ui/screen/onboarding/signin/SignInViewModel.kt
minhnimble: yeah let me refactor it in the next PR

PR #10, opened 2020-10-09
[TASK-5] [UI] As a user, I can see the home screen started with loading animation

minhnimble:
https://github.com/minhnimble/android-surveys-list/projects/1#card-46081790

## What happened

When users are logged in successfully, they will be redirected to the home dashboard with initial shimmer loading indicator.
 
## Insight

**Design**
[Shimmer loading animation](https://www.figma.com/proto/GjRPOjDyZ6f4EDL3wKarRK/Challenge---Mobile-App?node-id=34%3A49&scaling=scale-down)

**Acceptance Criteria**
- [x] Implement shimmer loading animations
 
## Proof Of Work

Screen Recording 2020-10-09 at 12 31 06


app/src/main/res/layout/activity_main.xml
```diff
@@ -18,6 +18,134 @@
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent"
         app:layout_constraintVertical_bias="0.0"
-        app:navGraph="@navigation/main_nav_graph" />
+        app:navGraph="@navigation/main_nav_graph" >
+
+    </androidx.fragment.app.FragmentContainerView>
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
manh-t, line 25: it would be so nice if we extract this layout into separate layout and `include` it here 🙏

manh-t (approved):
rest LGTM

app/src/main/res/layout/activity_main.xml
minhnimble: got it, nice suggest for code organizing, I will update 👍

app/src/main/res/layout/activity_main.xml
```diff
@@ -18,6 +18,17 @@
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent"
         app:layout_constraintVertical_bias="0.0"
```
hoangnguyen92dn, line 20: Why do we need `app:layout_constraintVertical_bias="0.0"`? 🤔

app/src/main/res/layout/activity_main.xml
```diff
@@ -18,6 +18,17 @@
...
+    </androidx.fragment.app.FragmentContainerView>
+
+    <include
+        layout="@layout/shimmer_loading_indicator_main"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent" />
```
hoangnguyen92dn, line 32: Because the layout always match parent as we defined from `@layout/shimmer_loading_indicator_main `, could we remove unnecessary attributes?

```
android:layout_width="match_parent"
        android:layout_height="match_parent"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toTopOf="parent"
```

app/src/main/res/layout/shimmer_loading_indicator_main.xml
```diff
@@ -0,0 +1,124 @@
...
+    android:background="@color/colorNero">
+
+    <com.facebook.shimmer.ShimmerFrameLayout
+        android:id="@+id/sflMainContainer"
+        android:layout_width="0dp"
+        android:layout_height="0dp"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent">
```
hoangnguyen92dn, line 16: Why don't we use `match_parent` and get rid of contains?

app/src/main/res/layout/shimmer_loading_indicator_main.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:id="@+id/clMainShimmerContainer"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:background="@color/colorNero">
```
hoangnguyen92dn, line 7: I'd suggest to use `com.facebook.shimmer.ShimmerFrameLayout` as parent for view rendering optimization. So we could get rid of ConstraintLayout, it's useless here.

app/src/main/res/layout/shimmer_loading_indicator_main.xml
```diff
@@ -0,0 +1,124 @@
...
+                android:id="@+id/clMainShimmerAvatar"
+                android:layout_width="@dimen/_28sdp"
+                android:layout_height="@dimen/_28sdp"
+                android:layout_marginEnd="@dimen/_15sdp"
+                android:background="@drawable/bg_main_shimmer_avatar"
+                app:layout_constraintBottom_toBottomOf="@+id/clMainShimmerBar2"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintWidth_percent="0.3">
+
+            </androidx.constraintlayout.widget.ConstraintLayout>
```
hoangnguyen92dn, line 33: We don't have child view in this layout, what do you think about using `View` rather than `ConstraintLayout`?

app/src/main/java/co/nimblehq/ui/screen/main/MainActivity.kt
```diff
@@ -17,5 +23,15 @@ class MainActivity : BaseActivity() {
...
+        toggleShimmerLoader(true)
+    }
+
+    override fun toggleShimmerLoader(shouldShow: Boolean) {
+        clMainShimmerContainer.visibility = if (shouldShow) View.VISIBLE else View.GONE
+        if (shouldShow) {
+            sflMainContainer.startShimmer()
+        } else {
+            sflMainContainer.stopShimmer()
+        }
```
hoangnguyen92dn, line 35: 🤔  Validate `shouldShow` twice seems redundant, not a big deal. You could ignore it.

app/src/main/res/layout/activity_main.xml
minhnimble: ohh this a redundancy, nice spot, will remove

app/src/main/res/layout/shimmer_loading_indicator_main.xml
minhnimble: hmm, let me check if it is feasible, since right now I want a layer that can position proportionally on the screen and `ConstraintLayout` can definitely achieve it

app/src/main/res/layout/shimmer_loading_indicator_main.xml
minhnimble: initially I did like what you said but we have some UI issue that when we start shimmer layer, it has a transparent overlapping issue with the underneath layer, and the shimmer color while it is animating is not as we originally expect. As a result, we need the `ConstraintLayout` underneath for proper animation. You can try to remove and see the problem happens.

app/src/main/res/layout/shimmer_loading_indicator_main.xml
minhnimble: oh okay, it is just my reference that I like using `layout_constraint`. This one is actually not need it, will update

app/src/main/res/layout/activity_main.xml
minhnimble: I like using `layout_constraint`. This one is actually not need it, will update

app/src/main/java/co/nimblehq/ui/screen/main/MainActivity.kt
minhnimble: yeah this way it saves one line of code so I apply. 👌I think it is ok

app/src/main/res/layout/activity_main.xml
minhnimble: We have these issues when not using constraints, need to have them:
Screen Shot 2020-10-12 at 09 20 00

app/src/main/res/layout/activity_main.xml
minhnimble: same issue here: https://github.com/minhnimble/android-surveys-list/pull/10#discussion_r503006725, we need the constraints 🙏

app/src/main/res/layout/activity_main.xml
hoangnguyen92dn: If you use `match_parent` with `constraint` parent. It would cause empty view on Samsung device :)

app/src/main/res/layout/shimmer_loading_indicator_main.xml
hoangnguyen92dn: I got the issue, and solution is set `android:background="@color/colorNero"` for `clMain` from `activity_main`. Could you please try it on your side?

app/src/main/res/layout/activity_main.xml
hoangnguyen92dn: @minhnimble Please try the solution I mentioned below to get rid of `ConstraintLayout` as parent and the issue would be resolved as well ;)

app/src/main/res/layout/activity_main.xml
```diff
@@ -17,7 +17,15 @@
...
+        app:navGraph="@navigation/main_nav_graph">
+
+    </androidx.fragment.app.FragmentContainerView>
+
+    <include
+        layout="@layout/shimmer_loading_indicator_main"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent" />
```
hoangnguyen92dn, line 29: Still need to add `layout_width/layout_height` and `constraint`?

app/src/main/res/layout/shimmer_loading_indicator_main.xml
minhnimble: oh you can do that because right now there is no contents in the `navHostContainer` in `activity_main.xml`, in the future we will have default contents in `surveys_fragment.xml`, and the issue will re-introduce since the nature of the shimmer layer is that we can look through it, we need to have that constraint layout for always having a fix background for the shimmer to act like an overlay correctly, not transparency.

app/src/main/res/layout/shimmer_loading_indicator_main.xml
minhnimble: same issue here: https://github.com/minhnimble/android-surveys-list/pull/10#discussion_r503006725, we need the constraints 🙏

app/src/main/res/layout/activity_main.xml
minhnimble: can you share more about your solution? is this the solution?
> match_parent and get rid of contains

If so I applied it already it it prompts the error as https://github.com/minhnimble/android-surveys-list/pull/10#discussion_r503006725

app/src/main/res/layout/activity_main.xml
minhnimble: yeah, if I remove the two constraints we will have this issue: https://github.com/minhnimble/android-surveys-list/pull/10#discussion_r503006725

app/src/main/res/layout/shimmer_loading_indicator_main.xml
hoangnguyen92dn: Ahha. Gotcha.

app/src/main/res/layout/activity_main.xml
hoangnguyen92dn: I think it could a wrong warning and we could ignore, switch to `Code` mode and no warning with `match_parent`, btw, I am ok with both, `constraint` or `match_parent`, but when you use `constraint` and expected the width and height match parent view, please notice to put `0dp` for width/height instead of `match_parent`

hoangnguyen92dn (approved):

PR #11, opened 2020-10-12
[TASK-6] [UI] As a user, I can see the initial surveys list populated

minhnimble:

## What happened

When entered home dashboard, users will see an initial list of surveys that they can navigate through

This PR is for creating the UI part of the surveys list screen 🛠🎨

Added a refactor for indenting all the build grades version's declarations.
 
## Insight

**Design**
[Surveys list UI](https://www.figma.com/proto/GjRPOjDyZ6f4EDL3wKarRK/Challenge---Mobile-App?node-id=30%3A46&scaling=scale-down)

**Acceptance Criteria**
- [x] Implement surveys list UI
 
## Proof Of Work

Screen Recording 2020-10-12 at 12 19 35


app/src/main/res/layout/item_surveys_pager_content.xml
```diff
@@ -0,0 +1,66 @@
...
+
+    <Button
+        android:id="@+id/btSurveysItemNext"
+        android:layout_width="@dimen/_44sdp"
+        android:layout_height="@dimen/_44sdp"
+        android:layout_marginEnd="@dimen/_15sdp"
+        android:background="@drawable/bg_surveys_item_next"
+        app:layout_constraintBottom_toBottomOf="@+id/tvSurveysItemDescription"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintTop_toTopOf="@+id/tvSurveysItemDescription" />
```
manh-t, line 65: I think we can move this button to the main view instead. How about your opinion? 😁

manh-t (approved):
rest LGTM

app/src/main/res/drawable/bg_surveys_item_next.xml
```diff
@@ -0,0 +1,16 @@
...
+        <corners android:radius="@dimen/_22sdp"/>
+        <solid android:color="@color/colorWhite"/>
+    </shape>
+</item>
+<item android:drawable="@drawable/ic_general_black_next"
+    android:bottom="@dimen/_13sdp"
+    android:left="@dimen/_16sdp"
+    android:right="@dimen/_16sdp"
+    android:top="@dimen/_13sdp" />
+</layer-list>
```
hoangnguyen92dn, line 16: Formatting please 🙏

app/src/main/res/drawable/bg_surveys_item_next.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+
```
hoangnguyen92dn, line 2: Redundant breakline

app/src/main/res/drawable/ic_general_black_next.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="12dp"
+    android:height="21dp"
```
hoangnguyen92dn, line 3: I see you are using `sdp` for some drawable. Could we update to apply `sdp` here for consistency?

app/src/main/res/layout/fragment_surveys.xml
```diff
@@ -1,6 +1,79 @@
...
     android:id="@+id/clSurveys"
     android:layout_width="match_parent"
-    android:layout_height="match_parent" />
+    android:layout_height="match_parent"
+    android:background="@color/colorNero">
+
+    <androidx.viewpager2.widget.ViewPager2
+        android:id="@+id/vpSurveys"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
```
hoangnguyen92dn, line 13: How about `0dp` here? As we already discussed about issue might caused on Samsung device?

app/src/main/res/layout/fragment_surveys.xml
```diff
@@ -1,6 +1,79 @@
...
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent" />
+
+    <me.relex.circleindicator.CircleIndicator3
+        android:id="@+id/ciSurveysIndicator"
+        android:layout_width="wrap_content"
+        android:layout_height="10dp"
```
hoangnguyen92dn, line 22: How about consider to use `sdp` here for consistency?

app/src/main/res/layout/fragment_surveys.xml
```diff
@@ -1,6 +1,79 @@
...
+        android:layout_marginBottom="@dimen/_159sdp"
+        app:ci_drawable="@drawable/ic_surveys_indicator_selected_circle"
+        app:ci_drawable_unselected="@drawable/ic_surveys_indicator_unselected_circle"
+        app:ci_height="5dp"
+        app:ci_width="5dp"
```
hoangnguyen92dn, line 27: Same here 🙏  It would be nice if we define `dimens` for reference, how do you think? And the design showing the size of indicator is `8` rather than `5` 🤔

app/src/main/res/layout/fragment_surveys.xml
```diff
@@ -1,6 +1,79 @@
...
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintHorizontal_bias="0.09"
```
hoangnguyen92dn, line 30: Following the design, we are not sure how much space will take with `0.09`, I think we could define the margin space following design on Figma? 🤔

app/src/main/res/layout/fragment_surveys.xml
```diff
@@ -1,6 +1,79 @@
...
+        android:id="@+id/tvSurveysToday"
+        style="@style/Widget.Template.TextView.Header"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="@dimen/_15sdp"
+        android:layout_marginEnd="@dimen/_8sdp"
+        android:text="@string/today"
+        app:layout_constraintEnd_toStartOf="@+id/cvSurveysUserAvatar"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="@+id/cvSurveysUserAvatar" />
```
hoangnguyen92dn, line 56: Following design, I guess it should be `bold`

app/src/main/res/layout/item_surveys_pager_content.xml
```diff
@@ -0,0 +1,66 @@
...
+
+    <TextView
+        android:id="@+id/tvSurveysItemHeader"
+        style="@style/Widget.Template.TextView.Header"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="@dimen/_15sdp"
+        android:layout_marginBottom="@dimen/_12sdp"
+        android:gravity="start"
+        android:maxLines="2"
```
hoangnguyen92dn, line 36: I think we need to add `ellipsize` to display `...` if the TextView doesn't fit the text length

app/src/main/res/values/widget_styles.xml
```diff
@@ -16,4 +16,18 @@
         <item name="android:focusable">true</item>
         <item name="android:background">?android:attr/selectableItemBackground</item>
     </style>
+
+    <style name="Widget.Template.TextView.Description">
+        <item name="android:textSize">@dimen/_13ssp</item>
+        <item name="android:letterSpacing">-0.02</item>
```
hoangnguyen92dn, line 22: 🤔 I guess you added here to get rid of default space from original font. In my opinion, we could ignore default space from the font we are using 😅

app/src/main/res/layout/item_surveys_pager_content.xml
minhnimble: oh I actually have a thought about this already, but when we do the horizontal swipe with the current design, the text will overlap with the button if we bring it on the main view. Do you have any suggestion for handling the swipe? 🤔

app/src/main/res/drawable/bg_surveys_item_next.xml
minhnimble: As of now I am still not sure what is the correct format for an xml file, can you please share me some info? 🙏

app/src/main/res/drawable/bg_surveys_item_next.xml
minhnimble: okay will remove

app/src/main/res/drawable/ic_general_black_next.xml
minhnimble: this one is a generated design from Figma, we should not modify it since every time we get the new designs from it we have to update again, which is not a good approach

app/src/main/res/layout/fragment_surveys.xml
minhnimble: okay 0dp is good, let me update

app/src/main/res/layout/fragment_surveys.xml
minhnimble: oh nice catch, let me update

app/src/main/res/layout/fragment_surveys.xml
minhnimble: I am okay with creating a dimens for this, let me check hmm

app/src/main/res/layout/fragment_surveys.xml
minhnimble: hmm, what if we have too many items and the indicators exceed the screen width? Trying to fit it into the screen with this approach

app/src/main/res/layout/fragment_surveys.xml
minhnimble: yeah you are right, let me update

app/src/main/res/values/widget_styles.xml
minhnimble: ah actually it is the style from Figma here 🥴, trying to make it look like design:

Screen Shot 2020-10-12 at 18 46 28

app/src/main/res/layout/item_surveys_pager_content.xml
minhnimble: oh you are right, let me update, nice catch

app/src/main/res/drawable/ic_general_black_next.xml
hoangnguyen92dn: Actually, `dp` is a default unit when you import svg file to AS. Because we are using `sdp` for better UI in different screen sizes, so we should apply for all dimens, otherwise, UI could be broken on some devices.

app/src/main/res/values/widget_styles.xml
hoangnguyen92dn: 👨‍🚀 In my opinion, we could set `includeFontPadding=false` and that's all. By using `android:letterSpacing`, it would reduce space between letter 🤔 We could ignore it for now 😅

app/src/main/res/layout/fragment_surveys.xml
hoangnguyen92dn: I think I didn't describe my idea clearly for you 😅 I meant you are using horizontal bias for indicators to anchor it on the left of screen but on different screens the space might be different, so that's why I suggest to use margin as design on Figma

image

app/src/main/res/drawable/bg_surveys_item_next.xml
hoangnguyen92dn: I am not sure about the style you applied to your AS. But as normally, <item> tags must be indented at least 4 spaces.

app/src/main/res/layout/fragment_surveys.xml
minhnimble: oh I got your point now, let me update 👌

app/src/main/res/drawable/ic_general_black_next.xml
minhnimble: I understand your point, let me try to find a solution for this

app/src/main/res/layout/item_surveys_pager_content.xml
manh-t: @minhnimble I'm thinking about the alternative approach for this, so we can remove ViewPager and use `TextSwitcher` and `ImageSwitcher` instead. Technically we can apply `fadeIn` and `fadeOut` animation in these Views, and it will look same as the design 🤔  Reference: https://medium.com/mobile-app-development-publication/switching-text-with-animation-f129eef83786

minhnimble: oh I think your idea is really great in terms of UX, the approach sounds a bit complicated for me but I would love come back for it after finalizing the functionality and logic for the application. Thank you so much for your idea 👍, btw I am not aware that we have `TextSwitcher` and `ImageSwitcher` 😓, still got a long way to learn Android 😂

manh-t: you're welcome, it's good to learn more ;) 🚀

app/src/main/res/layout/fragment_surveys.xml
hoangnguyen92dn: Seems you missed this comment.

app/src/main/res/layout/fragment_surveys.xml
```diff
@@ -1,6 +1,78 @@
...
+    <androidx.cardview.widget.CardView
+        android:id="@+id/cvSurveysUserAvatar"
+        android:layout_width="@dimen/_31sdp"
+        android:layout_height="@dimen/_31sdp"
+        android:layout_marginTop="@dimen/_60sdp"
+        android:layout_marginEnd="@dimen/_15sdp"
+        app:cardCornerRadius="@dimen/_15sdp"
+        app:cardElevation="0dp"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintTop_toTopOf="parent">
```
hoangnguyen92dn, line 66: Sorry I missed this from my previous review. Why do we need to wrap the avatar by cardview?

minhnimble: no worries, 👌 I need it for the corner radius

hoangnguyen92dn: 🤔 Following design on Figma, the avatar is a circle image. But CardView's shape is rectangle. Please correct me if I am wrong here.

minhnimble: we have the following attribute `app:cardCornerRadius="@dimen/_15sdp"` that already rounds the image view and create a circle container for whatever image view display inside 😓

hoangnguyen92dn: I don't think so. It's really hacky when using CardView with `app:cardCornerRadius` for circle image. Let's try with other drawable and you will see the issue. We need to implement crop image for cropping image to circle with any shape of images.

image

minhnimble: I don't understand your point, CardView with `app:cardCornerRadius` **CAN** create a circle container for any provided image. The screenshot you are showing is actually the preview of the ImageView's content and at this point it is not being rendered. 

Can you try to build the app with that `ic_launcher_background` and see the result. As I said, the CardView acts like a circle container for whatever image view's shape being displayed inside. I think there is no issue with this approach and force us to implement the extra crop image function.

Used this approach from this reference:
https://stackoverflow.com/questions/22105775/imageview-in-circular-through-xml

hoangnguyen92dn: @minhnimble I got your idea about using CardView with corner radius but as normally, the image inside is not really rounded circle, you can see the CardView will cut off the image a little bit. It's the simplest way but not the best way to cover all the cases.

image

luongvo: @minhnimble CardView should not be used for this case to produce a circle image view. CardView is made to produce a card view layout with shadow and rounded rectangle shape. In this case, we should consider applying a bitmap transformation such as [Glide transform](https://bumptech.github.io/glide/javadocs/400/com/bumptech/glide/load/resource/bitmap/CircleCrop.html). Rendering the layout by CardView and rendering a bitmap by Glide are differently, and Glide rendering should bring better performance.

app/src/main/res/layout/fragment_surveys.xml
```diff
@@ -1,6 +1,78 @@
...
+        android:layout_width="0dp"
+        android:layout_height="@dimen/_8sdp"
+        android:layout_marginStart="@dimen/_13sdp"
+        android:layout_marginBottom="@dimen/_150sdp"
+        app:ci_drawable="@drawable/ic_surveys_indicator_selected_circle"
+        app:ci_drawable_unselected="@drawable/ic_surveys_indicator_unselected_circle"
+        app:ci_height="@dimen/ci_size"
+        app:ci_width="@dimen/ci_size"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintStart_toStartOf="parent" />
```
hoangnguyen92dn, line 30: I think we should follow design to keep selected/unselected dot on indicators display with same size

image

sleepylee: it's the `app:ci_animator` that you may want to look at

app/src/main/res/layout/fragment_surveys.xml
sleepylee: please use the common practice with ImageView + image loader transformer, CardView has elevation and many other UI render support so even though it might work, it will cost more than we expect instead of using ImageView when rendering the image.

minhnimble: @hoangnguyen92dn, I want to answer your points as follow:
> the image inside is not really rounded circle, you can see the CardView will cut off the image a little bit

yeah no image is really rounded circle, and the `centerCrop` function actually do the same thing every time we perform the execution. The cut off happens because previously I set the CardView width and height is 31sdp but the cornerRadius is 15sdp and thus causing 1sdp cropping effect, not relating to the usage of CardView. This issue is fixed on the latest branch that the cornerRadius should be exactly half or the view's width and height.

> It's the simplest way but not the best way to cover all the cases.

not sure what corner cases you are talking about is which case. As of now, this is just a draft UI PR that load an image from resource, not in-cooperating logic into that view yet. When implementing logic and we have a need to load url and do cropping images via url, I think I will definitely consider using it, but not at the moment.

app/src/main/res/values/strings_no_translation.xml
```diff
@@ -1,4 +1,8 @@
 <resources>
-    <string name="sign_in_nimble_logo_desc">Nimble Logo</string>
     <string name="onboarding_splash_background_desc">Splash Screen Background</string>
+    <string name="sign_in_nimble_logo_desc">Nimble Logo</string>
+    <string name="user_avatar">User avatar</string>
+    <string name="surveys_sample_date_desc">MONDAY, JUNE 15</string>
+    <string name="surveys_sample_item_description">We would like to know how you feel about our work from home</string>
+    <string name="surveys_sample_item_header">Working from home Check-In</string>
```
sleepylee, line 7: for these resources please add `translatable="false"` to avoid the lint from AS, as been stated like the file name itself

app/src/main/res/values/strings.xml
```diff
@@ -9,4 +9,5 @@
     <string name="sign_in_password_hint">Password</string>
     <string name="sign_in_forgot_button_title">Forgot?</string>
     <string name="test_error">Test error</string>
+    <string name="today">Today</string>
```
sleepylee, line 12: for the general purpose, we have a convention to tie with `general_` if I'm not mistaken

app/src/main/res/values/strings_no_translation.xml
minhnimble: got it, will update

app/src/main/res/values/strings.xml
minhnimble: hmm, let me check and fix for the rest

app/src/main/res/layout/fragment_surveys.xml
minhnimble: oh right now it is not the same, let me take a look and fix it, thanks for spotting 👍

app/src/main/res/values/strings.xml
luongvo: `general_ ` or `common_` is okay

app/src/main/res/layout/fragment_surveys.xml
hoangnguyen92dn: @minhnimble 😅 I don't mention about using `centerCrop`, sorry if I made you confused about my idea. For common implementation and projects we are developing,  ImageView + Loader Transformation (As you might already known, Glide for loading image from URL) is the solution we often use.

Regarding to your solution by using CardView, I guess the `cornerRadius` must be a half of `width` or `height`, so it causes the issue when you use `31sdp` for width and height. As Lucas and Trung mentioned above, I  expected to see common practice from your implementation 😅

As you said `this is just a draft UI` so the ImageView with circle image from drawable for displaying is acceptable solution. I believe we don't need to wrap the ImageView by CardView as current.

minhnimble: @hoangnguyen92dn oh thanks for clarifying your point. Yup I was talking about `centerCrop` and `circleCrop` of Glide for image transformation. I guess it is something like this is what you mentioned right? 😊

Screen Shot 2020-10-13 at 17 48 54

oh okay I think CardView is not needed at the moment to be honest, will remove 👍

app/src/main/res/values/strings.xml
minhnimble: I am going with `general` @luongvo 😉

app/src/main/res/layout/fragment_surveys.xml
hoangnguyen92dn: @minhnimble That's comment from Lucas 😅

hoangnguyen92dn (approved):
LGTM

PR #12, opened 2020-10-13
[TASK-6] [Logic] As a user, I can see the initial surveys list populated

minhnimble:

## What happened

When entered home dashboard, users will see an initial list of surveys that they can navigate through

This PR is for implementing load initial surveys list page from backend
 
## Insight

**Acceptance Criteria**
- [x] Connect to surveys list API
- [x] Implement pager indicators list to be dynamic and based on the API response
- [x] Implement initial load surveys list logic if not cached

**Testing and implementation resources**
- Please make sure to use Staging environment for testing purposes, thanks 🙏
 
## Proof Of Work

Screen Recording 2020-10-13 at 15 17 11


app/src/main/java/co/nimblehq/data/model/Survey.kt
```diff
@@ -0,0 +1,31 @@
...
+    val title: String
+) {
+    val highResImageUrl: String
+        get() = imageUrl + "l"
+
+    constructor() : this(
+        "",
+        "",
+        "",
+        ""
```
sleepylee, line 18: you can put these to the default value of the attributes, normally we don't do like this for an empty constructor 🙏  its readability is kinda bad. This will bring better readability (you know which field has which kind of default value) and having the same result:
```
data class Survey(
    val description: String = "",
    val id: String = "",
    val imageUrl: String = "",
    val title: String = ""
)
```

app/src/main/java/co/nimblehq/data/model/Survey.kt
```diff
@@ -0,0 +1,31 @@
...
+    )
+}
+
+fun SurveysListingResponse.toSurveys() = with(data) {
+    this.map {
```
sleepylee, line 23: 🤔 do we really need to `map` after `with` and using `it`? 🤔  I haven't tried but can we just return sth like:

```
fun SurveysListingResponse.toSurveys() = with(data) {
    Survey(attributes.description, id, attributes.coverImageUrl, attributes.title)
}

File: app/src/main/java/co/nimblehq/data/model/Survey.kt [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+        Survey(
+            it.attributes.description,
+            it.id,
+            it.attributes.coverImageUrl,
+            it.attributes.title
```
sleepylee, line 28: at first I was a bit curious why do those fields are included in `attributes` 😁  it seems like then we are using http://jsonapi.org/ format like in The 1. You may want to recheck the Converter or the Response layer later. The Model/Entity is still safe now, later when you're reaching to Survey API, you will have a bit of a headache if you intend to map the fields manually

app/src/main/java/co/nimblehq/di/modules/RepositoryModule.kt
```diff
@@ -0,0 +1,30 @@
...
+@Module
+class RepositoryModule {
+
+    @Provides
+    @Singleton
+    fun provideAuthRepository(authService: AuthService): AuthRepository = ApiRepositoryProvider.getAuthRepository(authService)
+
+    @Provides
+    @Singleton
+    fun provideSurveyRepository(surveyService: SurveyService): SurveyRepository = ApiRepositoryProvider.getSurveyRepository(surveyService)
```
sleepylee, line 25: optional: I think you can remove the `ApiRepositoryProvider` and provide the implementation here. Since we are doing both at the `:app` module now, we might go straight to the implementation and reduce the memory footprint with `companion object` (static)

app/src/main/java/co/nimblehq/di/modules/RepositoryModule.kt
```diff
@@ -0,0 +1,30 @@
...
+import dagger.hilt.InstallIn
+import dagger.hilt.android.components.ApplicationComponent
+import javax.inject.Singleton
+
+@InstallIn(ApplicationComponent::class)
+@Module
+class RepositoryModule {
+
+    @Provides
+    @Singleton
```
sleepylee, line 20: I feel like the Auth will happen only at the login, then after that only when the token required to be refreshed then we need it again, so perhaps keep it Singleton always is a bit waste of memory? 🤔

data/src/main/java/co/nimblehq/data/api/providers/OkHttpClientProvider.kt
```diff
@@ -0,0 +1,20 @@
+package co.nimblehq.data.api.providers
+
+import co.nimblehq.data.BuildConfig
+import okhttp3.Interceptor
+import okhttp3.OkHttpClient
+
+class OkHttpClientProvider {
+    companion object {
+        fun getApiOkHttpClient(
```
sleepylee, line 9: suggesting to rename to `getOkHttpClient()` as we are inside the `OkHttpClientProvider` class, we should provide what we are stating

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
```diff
@@ -26,30 +30,45 @@ class SurveysFragment: BaseFragment(), BaseFragmentCallbacks {
...
-        tvSurveysDate.text = getString(R.string.surveys_sample_date_desc)
-        val testData = listOf(
-            SurveysPagerItemUiModel(R.drawable.bg_surveys_sample_item, getString(R.string.surveys_sample_item_header), getString(R.string.surveys_sample_item_description)),
-            SurveysPagerItemUiModel(R.drawable.bg_surveys_sample_item, getString(R.string.surveys_sample_item_header), getString(R.string.surveys_sample_item_description)),
-            SurveysPagerItemUiModel(R.drawable.bg_surveys_sample_item, getString(R.string.surveys_sample_item_header), getString(R.string.surveys_sample_item_description)),
-            SurveysPagerItemUiModel(R.drawable.bg_surveys_sample_item, getString(R.string.surveys_sample_item_header), getString(R.string.surveys_sample_item_description))
-        )
-
-        vpSurveys.adapter = SurveysPagerAdapter(testData).also {
+        tvSurveysDate.text = Date().getString(EEEE_MMMM_dd).toUpperCase()
```
sleepylee, line 38: getString(EEEE_MMMM_dd) <--- this is kinda vague and hard to understand at the first sight. I'd suggest to name it something closer to our needs and hide the format complexity away. maybe `Date().toDisplayFormat()` ?

data/src/main/java/co/nimblehq/data/api/providers/ApiServiceProvider.kt
```diff
@@ -1,12 +1,17 @@
 package co.nimblehq.data.api.providers
 
 import co.nimblehq.data.api.service.auth.AuthService
+import co.nimblehq.data.api.service.survey.SurveyService
 import retrofit2.Retrofit
 
 class ApiServiceProvider {
     companion object {
-        fun getAuthService(retrofit: Retrofit): AuthService {
+        fun getAuthService(retrofit: Retrofit) : AuthService {
```
sleepylee, line 9: I think the previous one is the correct format, this one when trying to run code reformat it should reject no? 🤔

data/src/main/java/co/nimblehq/data/lib/common/Const.kt
```diff
@@ -9,6 +10,18 @@ const val BLUR_SAMPLING = 2
...
 // Api
 
+const val AUTHORIZATION_HEADER = "Authorization"
+
 const val OAUTH_GRANT_TYPE_PASSWORD = "password"
 
 const val OAUTH_GRANT_TYPE_REFRESH_TOKEN = "refresh_token"
+
+internal const val REQUEST_KEY_PAGE_NUMBER = "page[number]"
+
```
sleepylee, line 20: Optional: I think we can remove this space to group these `REQUEST_KEY....` to the same group, not mandatory, it's up to the personal favor

data/src/main/java/co/nimblehq/data/lib/common/Const.kt
```diff
@@ -9,6 +10,18 @@ const val BLUR_SAMPLING = 2
...
+internal const val REQUEST_KEY_PAGE_SIZE = "page[size]"
+
+// Date
+
+const val dd_MMM_yyyy_hh_mm_a = "dd MMM yyyy, hh:mm a"
```
sleepylee, line 25: should name it something more readable, this snake_case_ violates the naming convention to us, also, the name is 90% as the value...something human friendly could be more appreciated: e.g: `DATE_FORMAT_FULL_DISPLAY` and the below can be: `DATE_FORMAT_SHORT_DISPLAY`

app/src/main/java/co/nimblehq/data/model/Survey.kt
minhnimble: oh I am totally not aware that we can do this, much cleaner, my goals is just to create an instance of this class with defaults value, so that is why I go with this, thanks for your suggestion 👍

app/src/main/java/co/nimblehq/data/model/Survey.kt
minhnimble: your proposed way doesn't work 😓

Screen Shot 2020-10-13 at 18 10 12

app/src/main/java/co/nimblehq/data/model/Survey.kt
minhnimble: noted, will definitely take a closer look into the layers you mentioned. Thanks for letting me know about that 👌

app/src/main/java/co/nimblehq/di/modules/RepositoryModule.kt
minhnimble: Got it, I initially create a provider layer for experimenting ways to provide two different retrofit instances for these services but forget to cleanup. Will clear them out, thanks 👍

app/src/main/java/co/nimblehq/di/modules/RepositoryModule.kt
minhnimble: hmm, in the case user logins and logouts a lot, it would be another story I think. 🤔 I think we should keep the Singleton here as it is until it actually make an impact to memory. What do you think? 🙏

data/src/main/java/co/nimblehq/data/api/providers/OkHttpClientProvider.kt
minhnimble: got it, will update

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
minhnimble: okay, let me update the function name to be more meaningful 👌

data/src/main/java/co/nimblehq/data/api/providers/ApiServiceProvider.kt
minhnimble: On my side it is not rejecting... hm I am not sure about this convention here, let me revert then

data/src/main/java/co/nimblehq/data/lib/common/Const.kt
minhnimble: How about I will create another section for this called `// Request key`. What do you think? 😁

data/src/main/java/co/nimblehq/data/lib/common/Const.kt
minhnimble: Actually this is something I intentionally name like this, this way I don't have to navigate to the definition to know what is the value of the date format. How about adding a `DATE_FORMAT_...` prefix for these strings for some clarity, what do you think? 🤔

data/src/main/java/co/nimblehq/data/lib/common/Const.kt
```diff
@@ -1,14 +1,30 @@
...
 
 const val OAUTH_GRANT_TYPE_PASSWORD = "password"
 
 const val OAUTH_GRANT_TYPE_REFRESH_TOKEN = "refresh_token"
+
+// Date formats
+
+const val DATE_FORMAT_dd_MMM_yyyy_hh_mm_a = "dd MMM yyyy, hh:mm a"
+
+const val DATE_FORMAT_EEEE_MMMM_dd = "EEEE, MMMM dd"
```
hoangnguyen92dn, line 23: Could we make all of letters in constant name uppercase?

And I'd suggest to have meaningful name for date time patterns, just wondering about what will you naming this pattern `EEEE MMMM dd` or `EEEE MMMM, dd` 😅

app/src/main/java/co/nimblehq/extension/DateExtension.kt
```diff
@@ -0,0 +1,10 @@
+package co.nimblehq.extension
+
+import co.nimblehq.data.lib.common.DATE_FORMAT_dd_MMM_yyyy_hh_mm_a
+import java.text.SimpleDateFormat
+import java.util.*
+
+fun Date.toDisplayFormat(format: String = DATE_FORMAT_dd_MMM_yyyy_hh_mm_a): String {
+    val sdf = SimpleDateFormat(format)
```
hoangnguyen92dn, line 8: It would better to add default Locale to get rid of warning.

`To get local formatting use getDateInstance(), getDateTimeInstance(), or getTimeInstance(), or use new SimpleDateFormat(String template, Locale locale) with for example Locale.US for ASCII dates.`

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
```diff
@@ -0,0 +1,30 @@
...
+import co.nimblehq.data.api.service.survey.SurveyService
+import co.nimblehq.data.model.Survey
+import co.nimblehq.data.model.toSurveys
+import io.reactivex.Single
+import javax.inject.Inject
+
+interface SurveyRepository {
+
+    fun getSurveysList(
+        pageNumber: Int = 0,
```
hoangnguyen92dn, line 12: Just curious about default `pageNumber = 0` here, as I can see we already defined default = 1 from ViewModel 🤔

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -2,14 +2,44 @@ package co.nimblehq.ui.screen.main.surveys
...
+
+    val surveysPagerItemUiModels: Observable<List<SurveysPagerItemUiModel>>
+        get() = _surveysPagerItemUiModels
+
+    fun getSurveysList(pageNumber: Int = 1, pageSize: Int = 10) {
+        getSurveysListSingleUseCase
+            .execute(GetSurveysListSingleUseCase.Input(pageNumber, pageSize))
+            .doFinally { _showLoading.onNext(false) }
+            .subscribeBy(
+                onSuccess = { _surveysPagerItemUiModels.onNext(it.map { survey -> survey.toSurveysPagerItemUiModel() }) },
```
hoangnguyen92dn, line 39: Imo, we could `map` the survey before emitting will make the stream looks better. You could keep it. I am ok with both 😅

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -2,14 +2,44 @@ package co.nimblehq.ui.screen.main.surveys
...
+    val showLoading: Observable<Boolean>
+        get() = _showLoading
+
+    val surveysPagerItemUiModels: Observable<List<SurveysPagerItemUiModel>>
+        get() = _surveysPagerItemUiModels
+
+    fun getSurveysList(pageNumber: Int = 1, pageSize: Int = 10) {
+        getSurveysListSingleUseCase
+            .execute(GetSurveysListSingleUseCase.Input(pageNumber, pageSize))
+            .doFinally { _showLoading.onNext(false) }
```
hoangnguyen92dn, line 37: Can you please point me to where will we do `_showLoading .onNext(true)`? 🤔

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
```diff
@@ -26,30 +30,45 @@ class SurveysFragment: BaseFragment(), BaseFragmentCallbacks {
...
+
+        viewModel.surveysPagerItemUiModels
+            .subscribe(::bindSurveysPagerItemUiModels)
+            .bindForDisposable()
+    }
+
+    private fun bindLoading(isLoading: Boolean) {
+        if (!isLoading)
+            loaderAnimator?.toggleShimmerLoader(false)
+        toggleLoading(isLoading)
```
hoangnguyen92dn, line 66: It would be nice if we have brackets here 😅

data/src/main/java/co/nimblehq/data/lib/common/Const.kt
minhnimble: Hi @hoangnguyen92dn, I would like to answer your points as follow:
> Could we make all of letters in constant name uppercase?

This is quite a unique situation since the date format with upper char and lower char has different meaning, i.e. `DD` and `dd`. So it kinda defeats the purpose of looking at the key representation and we can know the value right away. 

> And I'd suggest to have meaningful name for date time patterns, just wondering about what will you naming this pattern EEEE MMMM dd or EEEE MMMM, dd 😅

Looking at your example, I thought that this case we can add a number postfix like `DATE_FORMAT_EEEE_MMMM_dd_1` and `DATE_FORMAT_EEEE_MMMM_dd_2` but in the end, we will have to check the definition anyway. As a result, I guess I will go with Trung's suggestion 🥴. Thanks for bringing this up 🙏

app/src/main/java/co/nimblehq/extension/DateExtension.kt
minhnimble: oh I don't have any warnings, let me check again and fix it, thanks 🙏

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
minhnimble: hmm, I tested the API and noticed that providing 0 for pageNumber will result in a `Server internal error: 500`. So I updated the ViewModel but not here, let me update here as well 👌

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: hmmm, in the future I will modify this part to support local caching, for now you can ignore here 😉

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: in the future we will have load more capability, `_showLoading .onNext(true)` will be used at that time. At the moment, you can ignore 👌

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
minhnimble: I think it is just a one line check and it is too nitpicking for adding brackets here, I would prefer to add the brackets when we have more than one line pieces of code. how about making this if part to be in one-lined, what do you think?

data/src/main/java/co/nimblehq/data/lib/common/Const.kt
minhnimble: After further discussion with @hoangnguyen92dn here https://github.com/minhnimble/android-surveys-list/pull/12#discussion_r504359939, I think it is much better to go with your suggestion, will update 👍

app/src/main/java/co/nimblehq/di/modules/RepositoryModule.kt
sleepylee: optimization can come later sure 👍

data/src/main/java/co/nimblehq/data/api/providers/ApiServiceProvider.kt
sleepylee: when in doubt, just try to make sure that you're using the default setting 🙏 
image

Notice the function return type will not have space, but the class inheritance will have 😁 
image

data/src/main/java/co/nimblehq/data/lib/common/Const.kt
sleepylee: we don't need the comment if you're grouping them in a meaningful way like that 😁  imagine coming to a code base like this:

```
internal const val REQUEST_KEY_PAGE_NUMBER = "test"
internal const val REQUEST_KEY_TEST = "test"
internal const val REQUEST_KEY_TEST_2 = "test"
internal const val REQUEST_KEY_TEST_3 = "test"
internal const val REQUEST_KEY_PAGE = "test"
```

you should be able to understand they are the `REQUEST_KEY` group right? 😄

data/src/main/java/co/nimblehq/data/lib/common/Const.kt
sleepylee: I think depending on the name, the idea is that when you want to just simply display a Date field in the app, you just want to use something like: `DEFAULT_DATE_FORMAT`, and then don't care about whatever shows up. When there is an issue comes, that might be the only time when you want to inspect the detail of that implementation 😁

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
sleepylee: or you can just include in the useCase 🤓

app/src/main/java/co/nimblehq/data/model/Survey.kt
sleepylee: I see, it's mapping each item from the list not a single data type..hmm...I think you may have to spend some research time to work with the converter as I mentioned below...since you're already at the stage where you need to communicate for the Survey API, the sooner you figure out the converter, the better it should be. 
Basically, the converter should flatten the json object for you, if you are looking for only data within the `SurveyResponse` it should contain all those fields: id, title, description at the top access level 🆗

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
sleepylee: 1 line like now is ok 👌 , but if we break into 2 lines `if` and then `action` then we must have braces, just a reminder 😁

sleepylee: here is a Fun read about putting action under if statement without braces 😂 
https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/

data/src/main/java/co/nimblehq/data/lib/common/Const.kt
minhnimble: yeah it is straight forward, let me update

app/src/main/java/co/nimblehq/data/model/Survey.kt
minhnimble: got your point, let me investigate about it

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: yup, will consider that in the future 👍

data/src/main/java/co/nimblehq/data/lib/common/Const.kt
hoangnguyen92dn: Yup. Got your point. My idea is just give the feedback as we are developing our current projects for client, so you could aware of what are the rules, conventions we are following on Android projects 😅  Basically, we should follow rules, conventions when we are working in the team, so I am ok with your ideas for personal challenge project 😅

app/src/main/java/co/nimblehq/data/model/Survey.kt
minhnimble: Implemented on #592a73c, please help me take a look

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: Updated as @hoangnguyen92dn comment in this commit: 5a9bfba 😉

manh-t (approved):

PR #13, opened 2020-10-15
[TASK-6] [Database] As a user, I can see the initial surveys list populated

minhnimble:

## What happened

When entered home dashboard, users will see an initial list of surveys that they can navigate through. 

This PR is for implementing surveys list caching after retrieving it from backend with Room Database. 
If the data is loaded in previous app session, then users terminate the app and reopen it, the app will use local cached data to display and at the same time trigger the API call to refresh surveys list data in the background.
 
## Insight

**Acceptance Criteria**
- [x] Implement cached logic to phone after loading initial surveys list

**Testing and implementation resources**
- Please make sure to use Staging environment for testing purposes, thanks 🙏
 
## Proof Of Work

Screen Recording 2020-10-15 at 13 40 30


app/src/main/java/co/nimblehq/data/model/Survey.kt
```diff
@@ -1,13 +1,22 @@
...
 
+import android.os.Parcelable
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.PrimaryKey
 import co.nimblehq.data.api.response.survey.SurveyResponse
+import kotlinx.android.parcel.Parcelize
 
+@Entity(tableName = "survey")
+
```
manh-t, line 11: redundant space? 🍏

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -33,12 +35,34 @@ class SurveysViewModel @ViewModelInject constructor(
...
-    fun getSurveysList(pageNumber: Int = DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER, pageSize: Int = DEFAULT_SURVEYS_PAGE_SIZE) {
-        getSurveysListSingleUseCase
-            .execute(GetSurveysListSingleUseCase.Input(pageNumber, pageSize))
+    fun checkAndLoadSurveysListIfNeeded() {
+        getLocalSurveysListSingleUseCase
+            .execute(Unit)
+            .map { it.map { survey -> survey.toSurveysPagerItemUiModel() } }
+            .subscribeBy(
+                onSuccess = {
+                    if (it.isEmpty()) {
```
manh-t, line 44: @minhnimble I wonder that do we need to handle in case the data is out of date? 🙏

minhnimble: @manh-t nice question, as per from the Andoird IC requirements, we will have the capability to refresh cache with pull to-refresh capability to invalidate the local cache, apart from that there is no other mentioning about when the local data is out of date:

`The cache of the list of surveys can be refreshed from the home screen by doing a pull-to-refresh.`

As a result, once the retrieved data is stored to local, it will be assumed to never go out of date until user triggers a pull-to-refresh action. This feature will be implemented in the [Task 7](https://github.com/minhnimble/android-surveys-list/projects/1#card-46082926). Does this answer your question? 😊

app/src/main/java/co/nimblehq/data/model/Survey.kt
minhnimble: ohhh, I thought we need, I saw there is a space in The1 code, will remove 👍

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
manh-t: image
@minhnimble oh, but from my understanding about the requirement, the surveys **must be fetched when opening the application**, so it means that we always have the newest content whenever the application launch. Please correct me if I'm wrong 🙇

minhnimble: @manh-t I think we can trigger the local RoomDB check and refresh the content at the same time in the background. Initially I just thought that either we trigger the API if there is no cache data or use the cache if we have it. Thanks for notifying, I will update it to this logic. Sounds good to you? 👌

manh-t: @minhnimble cool, look good to me.
I have a useful tip to refresh content in Room database with Rxjava2, then the new data will automatically emit when there is change in database. Hope it helps you minimize the implementation 🙏  Reference: https://medium.com/androiddevelopers/room-rxjava-acb0cd4f3757

minhnimble: Hi @manh-t, I implemented this logic in this commit d7e351a. Please help me take a look 🙏

manh-t (approved):
rest LGTM. Would you mind updating the PR's description to match with the new implementation? 🙏  😁

minhnimble:
> rest LGTM. Would you mind updating the PR's description to match with the new implementation? 🙏 😁

updated, nice spot 👍

PR #16, opened 2020-10-19
[TASK-7] [Logic] As a user, I can load more or refresh surveys list

minhnimble:
https://github.com/minhnimble/android-surveys-list/projects/1#card-46082926

## What happened

When users are looking at the surveys list, they will be able to scroll to the end of the list to load more or pull down to refresh the list.

This PR is for handling load more items logic when user scrolls and reaches to the end (right most) and refresh items logic when pulling down to refresh. Also, update UI to remove ViewPager, replace with animated TextView and ImageView and set the Next Button to be static on the screen to match with design. I cannot use the ImageSwitcher and TextSwitcher due to the fact that it cannot be constrained properly with Constraint Layout. Need to investigate further why but for now I will use the Animator instead to achieve the same behavior. 🙏
 
## Insight

- [x] Implement pagination logic to load more surveys
- [x] Implement caching logic after loading more surveys
- [x] Implement pull down to refresh logic to refresh the contents and load the new initial surveys list page
- [x] Implement clear cache logic when refreshing the contents
- [x] Extend surveys list API to support pagination
- [x] Use animated TextView and ImageView and static Next Button on screen.
 
## Proof Of Work

Load more logic:
Screen Recording 2020-10-19 at 18 33 47

Pull down to refresh logic:
Screen Recording 2020-10-20 at 15 22 35


app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,209 @@
...
+                        _selectedSurveyIndex.onNext(nextIndex)
+                        _selectedSurvey.onNext(surveys[nextIndex])
+                    }
+
+                    // When we reach the last item, trigger to load more
+                    if (surveys.size - 1 == nextIndex && !loadedLastSurveysPage || surveys.size == nextIndex) {
+                        loadMoreSurveys()
+                    }
+                }
+            }
```
manh-t, line 86: Too many conditions here 😅. At the first look, I don't see `else` condition for `when{}`, so does it mean that we only have 2 cases for this? If so, could we apply `if (it == DEFAULT_UNSELECTED_INDEX) {} else {}` here? 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,209 @@
...
+        }
+    }
+
+    fun refreshSurveysList() {
+        resetSurveysPaginationAttributes()
+
+        deleteLocalSurveysCompletableUseCase
+            .execute(Unit)
+            .doOnSubscribe { _showLoading.onNext(true) }
+            .andThen(loadMoreSurveysSingleUseCase.execute(LoadMoreSurveysSingleUseCase.Input(currentSurveysPageNumber, currentSurveysPageSize)))
```
manh-t, line 106: @minhnimble oh, I think there is the mismatch here. If we have `getInitialSurveysFlowableUseCase`, should we call it instead of `loadMoreSurveysSingleUseCase` here? 🤔

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,209 @@
...
     }
+
+    private fun bindOnErrorLoadSurveys(error: Throwable) {
+        if ((error as? SurveyError.GetSurveysError)?.isNotFound == true) {
+            loadedLastSurveysPage = true
+            _error.onNext(SurveyError.NoMoreSurveysError(null))
+        } else {
+            _error.onNext(error)
+        }
+    }
```
manh-t, line 158: I notice that we have `pages` returned value from API, not sure if it's the total page. If it is, we can set limit for load more instead of waiting to throw the error like this. Would you mind confirming with API team about this value? 🙏 Then we could avoid fetching `404: Page not found` 😂

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
```diff
@@ -31,20 +31,36 @@ class SurveysFragment: BaseFragment(), BaseFragmentCallbacks {
...
-        ciSurveysIndicator.setViewPager(vpSurveys)
     }
 
-    override fun bindViewEvents() { }
+    override fun bindViewEvents() {
+        btSurveysItemNext.subscribeOnClick {
+            // TODO: Handle navigate to survey details screen
+        }.bindForDisposable()
+
+        vSurveysItemTransparent.setOnTouchListener(object : OnSwipeTouchListener(requireContext()) {
```
manh-t, line 46: I see that you are creating a transparent view to handle swipe, how about if we use root view for this case? 🙏 Perhaps it is `clSurveys`

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: there are 3 cases here, `it == DEFAULT_UNSELECTED_INDEX` (first block), `it > DEFAULT_UNSELECTED_INDEX` (second block), `it < DEFAULT_UNSELECTED_INDEX` (we do not exec anything). I can refactor it to `if else` but the idea is the same, what do you think? 🤔

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: @manh-t, I actually have thought of that, however at this place `getInitialSurveysFlowableUseCase` will do one extra thing, which is getting local data from Room up again, but since we just delete everything in `deleteLocalSurveysCompletableUseCase`, there is no point in doing that and that is why I have to go with `loadMoreSurveysSingleUseCase`. What do you think? 🤔

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
minhnimble: okay let me try and see if it works, thanks for your suggestion 👍

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: oh right now when triggering load more, I have a flag to check when to trigger the next api call to server, but it has to trigger the last call and expect the 404 error to update the flag. I think your approach is good because it will save one API call, I will work on it 👍

minhnimble:
Hi @manh-t, I fixed and reply all your comments via 13d0d52. Please help me take a look again.

@hoangnguyen92dn, @sleepylee, @luongvo please help me review this PR as well 🙏

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
```diff
@@ -31,8 +38,20 @@ class SurveyRepositoryImpl @Inject constructor(
...
+    override fun loadMoreSurveysList(
+        pageNumber: Int,
+        pageSize: Int
+    ): Single<List<Survey>> {
+        return surveyService
+            .getSurveysList(pageNumber, pageSize)
+            .firstOrError()
+            .map { it.surveys.toSurveys() }
+            .doOnSuccess { surveyDao.insertAll(it) }
+    }
```
manh-t, line 56: Hi @minhnimble , firstly I see that you use one api endpoint for 2 places with the same purpose. Any reason why do you need to duplicate it here? 🙏 
And, I notice that you are using 2 sources for generating survey list, they are `network` and `database`. Should we apply single source of truth to get new data from only one source? I think it is `database` in this case, we will fetch data from network, then inserting data to the database, and we can listen the change from database and update the UI accordingly. How do you think?

app/src/main/java/co/nimblehq/data/storage/AppPreferences.kt
```diff
@@ -2,16 +2,25 @@ package co.nimblehq.data.storage
...
+import io.reactivex.Completable
 import javax.inject.Inject
 
 interface AppPreferences {
-    // TODO: Add new keys here
+    var surveysTotalPages: Int?
+
+    fun clearAll() = Completable.fromAction {
+        surveysTotalPages = null
+    }
```
manh-t, line 13: does it make sense to use `Completable` here? 🤔

app/src/main/java/co/nimblehq/ui/common/listener/OnSwipeTouchListener.kt
```diff
@@ -0,0 +1,68 @@
...
+
+open class OnSwipeTouchListener(ctx: Context) : View.OnTouchListener {
+
+    private val gestureDetector: GestureDetector
+
+    companion object {
+
+        private val SWIPE_THRESHOLD = 100
+        private val SWIPE_VELOCITY_THRESHOLD = 100
+    }
```
hoangnguyen92dn, line 16: Can we get rid of `companion object`? As we already define `const` in other classes.

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
```diff
@@ -55,8 +72,16 @@ class SurveysFragment: BaseFragment(), BaseFragmentCallbacks {
...
 
-        viewModel.surveysPagerItemUiModels
-            .subscribe(::bindSurveysPagerItemUiModels)
+        viewModel.selectedSurvey
+            .subscribe(::bindSelectedSurvey)
+            .bindForDisposable()
+
+        viewModel.selectedSurveyIndex
+            .subscribe(::bindSelectedSurveyIndex)
+            .bindForDisposable()
```
hoangnguyen92dn, line 81: I guess we could combine `selectedSurvey` and `selectedSurveyIndex`.

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,193 @@
...
+        _selectedSurveyIndex.value?.let {
+            val surveys = _surveyItemUiModels.value.orEmpty()
+            if (it == DEFAULT_UNSELECTED_INDEX) { // First time update selected index
+                setDisplaySurveyItemWith(0, surveys)
+            } else if (it > DEFAULT_UNSELECTED_INDEX) { // Selected index is updated before
+                val nextIndex = it + 1
+                setDisplaySurveyItemWith(nextIndex, surveys)
+                if (surveys.size - 1 == nextIndex && !reachedLastSurveysPage || surveys.size == nextIndex) { // When we reach the last item, trigger to load more
+                    checkAndLoadMoreSurveysIfNeeded()
+                }
```
hoangnguyen92dn, line 76: It would be better if we have threshold for loading more before reached the last item.

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,193 @@
...
+
+    override fun previousIndex() {
+        _selectedSurveyIndex.value?.let {
+            val surveys = _surveyItemUiModels.value.orEmpty()
+            val previousIndex = it - 1
+            setDisplaySurveyItemWith(previousIndex, surveys)
+        }
+    }
+
+    fun refreshSurveysList() {
```
hoangnguyen92dn, line 89: Could we reset value in SharedPreferences at this step as well? 🤔

hoangnguyen92dn (commented):
Hi @minhnimble, sorry for the late review on this PR. As the requirements, I think we just need to cache survey list in in-app memory rather than database. If we store in the database, could we display survey to UI in case no connectivity? And as the quick look, there are some unused classes and imports (DiffUpdateAdapter, RetrofitModule), can you check again to make sure clean unnecessary things from your code base.

app/src/main/java/co/nimblehq/data/storage/AppPreferences.kt
minhnimble: yeah, in the future I will create a LogoutCompletableUsecase  and I will use this function to clean up sharedPreferences 🙏

app/src/main/java/co/nimblehq/data/error/SurveyError.kt
```diff
@@ -8,8 +8,19 @@ sealed class SurveyError(
     @StringRes readableMessageRes: Int? = null
 ) : AppError(cause, (cause as? JsonApiException)?.error?.detail, readableMessageRes) {
     
-    class GetSurveysListError(cause: Throwable?) : SurveyError(
-        cause, R.string.general_get_surveys_list_error
+    class GetSurveysError(cause: Throwable?) : SurveyError(
+        cause, R.string.general_get_surveys_error
+    ) {
+        val isNotFound: Boolean
+            get() = code == 404
```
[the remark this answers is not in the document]
minhnimble: oh I forget to clean up this val since I updated the code to use pages value instead of keep calling request and use `HTTP_NOT_FOUND` exception as the termination condition for pagination loading. Will remove 👍

app/src/main/java/co/nimblehq/data/storage/AppPreferences.kt
```diff
@@ -2,16 +2,25 @@ package co.nimblehq.data.storage
...
 }
 
 class AppPreferencesImpl @Inject constructor(
     private val preferences: SharedPreferences
 ) : AppPreferences {
 
+    override var surveysTotalPages: Int?
+        get() = getInt(KEY_SURVEYS_TOTAL_PAGES)
+        set(theme) = setOrRemove(KEY_SURVEYS_TOTAL_PAGES, theme)
```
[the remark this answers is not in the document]
minhnimble: oh wrong naming, let me update 😅

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
minhnimble: Hi @manh-t, for the first endpoint, this technique is recommended by @sleepylee using The1 Android example that when we initially open the app, it will update the UI with local `database` first, at the same time trigger the `network` to update it at a later time when the `network` data arrives, this way we can populate data to users faster if it is already stored to database, not always having to wait the initial `network` data call like your approach. Also, using your approach, the more data in `database` source, every new page call we get the total data got bigger and bigger.

The second api endpoint is for another needs that during load more use case, there is no local `database` source when triggering it, and thus we just need the endpoint to get new `network` data page, store it to `database` and send that same page of data to the use cases. I can reuse the first api endpoint but it will just redundantly giving me the `database` source, which is not really needed. 

What do you think? Maybe I can try to update the first endpoint logic in the usecase layer? 🤔 Does it sounds good to you?

app/src/main/java/co/nimblehq/data/error/AppError.kt
```diff
@@ -1,15 +1,16 @@
...
 open class AppError(
     cause: Throwable?,
     open val readableMessage: String? = null,
     @StringRes open val readableMessageRes: Int? = null
 ) : Throwable(cause) {
 
-    protected open val code: String?
-        get() = (cause as? JsonApiException)?.error?.code
+    protected open val code: Int?
+        get() = (cause as? HttpException)?.code()
```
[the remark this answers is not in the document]
minhnimble: Okay let me take a look, previously I was trying to get the code from `JsonApiException` without via `JsonApiException.response` but using direct `JsonApiException` property and it always give me the empty string so I have to do this.

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
```diff
@@ -4,25 +4,32 @@ import co.nimblehq.data.api.service.survey.SurveyService
...
+import co.nimblehq.data.storage.AppPreferences
 import co.nimblehq.data.storage.dao.SurveyDao
 import io.reactivex.Flowable
 import io.reactivex.Single
 import javax.inject.Inject
 
 interface SurveyRepository {
 
-    fun getSurveysList(
+    fun getInitialSurveysListWithCache(
```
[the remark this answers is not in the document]
minhnimble: let me try to put it in the usecase layer, I was using references from The 1 but it is indeed causing repeated code here 🙏

app/src/main/java/co/nimblehq/data/storage/AppPreferences.kt
```diff
@@ -2,16 +2,25 @@ package co.nimblehq.data.storage
 
 import android.content.SharedPreferences
 import androidx.core.content.edit
+import io.reactivex.Completable
 import javax.inject.Inject
 
 interface AppPreferences {
-    // TODO: Add new keys here
+    var surveysTotalPages: Int?
```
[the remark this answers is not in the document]
minhnimble: yeah, we need it in the case we have loaded a few pages of data in the database already, then from there opening the app without any connection.
As soon as we have the connection back and we trigger load more action by swiping all the way to the right, we would know if we have loaded everything or not and don't need to trigger another api call just to check as per suggestion from @manh-t

app/src/main/java/co/nimblehq/ui/common/listener/OnSwipeTouchListener.kt
minhnimble: noted, will update

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
minhnimble: oh interesting point, let me try

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: oh I have already set a threshold check inside `checkAndLoadMoreSurveysIfNeeded` as the function name stated. 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: oh nice catch, will update

minhnimble: actually, we will reset everything from the beginning so I will also reset `SharedPreferences`here 👍

minhnimble:
Hi @manh-t and @hoangnguyen92dn, thanks so much for your reviews and suggestions, I updated everything here 6b3692a, code also looks much cleaner, please help me take a look

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
hoangnguyen92dn: I mean threshold for preload next page rather than start loading when user reached last item on current page. For ex:
Page1: 10 items -> When user reaches item 8th, we could start loading next page.

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
manh-t: @minhnimble Sorry if I haven't explained to you clearly but it didn't what I mean. 🙏 

> when we initially open the app, it will update the UI with local database first, at the same time trigger the network to update it at a later time when the network data arrives, this way we can populate data to users faster if it is already stored to database

It is okay, but when I open app in case no internet connection -> The app is failed to load, I think you need to handle this case.

> not always having to wait the initial network data call like your approach.

I don't understand what you mean here. My approach is to prioritize data from `database` first if it has, then will trigger network call to fetch the newest data, like your approach when calling from `network` as well, but it is `database` in my case

> Also, using your approach, the more data in database source, every new page call we get the total data got bigger and bigger.

The data in database will be replaced with newer data, what does it mean `every new page call we get the total data got bigger and bigger.`? 🙏

minhnimble:
> ...I think we just need to cache survey list in in-app memory rather than database. If we store in the database, could we display survey to UI in case no connectivity

Yeah right now if you loaded some data in previous session and restart the app without connection, it will display cache data. Not sure why you just want to just cache survey list in in-app memory? 🤔 Caching in database will provide a much better users experience... 🧐

> And as the quick look, there are some unused classes and imports (DiffUpdateAdapter, RetrofitModule), can you check again to make sure clean unnecessary things from your code base.
 
Scanned and updated, thanks for reminding me about this 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,196 @@
...
+interface Inputs {
+
+    fun nextIndex()
+
+    fun previousIndex()
+}
 
 class SurveysViewModel @ViewModelInject constructor(
-    private val getSurveysListFlowableUseCase: GetSurveysListFlowableUseCase
+    private val appPreferences: AppPreferences,
```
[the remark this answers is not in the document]
minhnimble: sure, let me refactor it here 👍

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: oh I understand your point now. Currently, the threshold is when user reaches the last item (10th item), and since we are not using a recycle view that have multiple items in one screen, but instead a single item per screen,
I don't think it is necessary for changing the threshold to 8th for preloading next page. 

The goal for your request is to make the scroll smooth when showing multiple items in a single screen but as of now I think it is smooth enough since we are using the full screen per item design.

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
minhnimble: > It is okay, but when I open app in case no internet connection -> The app is failed to load, I think you need to handle this case.

Totally agree with you that we will need to handle no internet connection case. However, there is no requirement in the Android Certification for doing it so I will implement it in a separate PR when the app is fully functional as an addition optimization. Sounds good to you? 😊

>I don't understand what you mean here. My approach is to prioritize data from database first if it has, then will trigger network call to fetch the newest data, like your approach when calling from network as well, but it is database in my case

oh I guess you want to keep the first api endpoint since that is the approach you mention, any how, I updated it to move the prioritizing data from `database` first to usecase layer. It is updated now, please help me take a look again 🙏

>The data in database will be replaced with newer data, what does it mean every new page call we get the total data got bigger and bigger.? 🙏

Previously I thought that your flow is trigger the network call to load 1 page -> save to database -> getting all data from database from this sentence `I think it is database in this case, we will fetch data from network, then inserting data to the database, and we can listen the change from database and update the UI accordingly.`. If that is not your point then we can just ignore 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,202 @@
...
     val error: Observable<Throwable>
         get() = _error
 
     val showLoading: Observable<Boolean>
         get() = _showLoading
 
-    val surveysPagerItemUiModels: Observable<List<SurveysPagerItemUiModel>>
-        get() = _surveysPagerItemUiModels
+    val selectedSurveyData: Observable<Pair<Int, SurveyItemUiModel?>>
+        get() = _selectedSurveyData
```
hoangnguyen92dn, line 46: From `selectedIndex`, I believe we are able to get current Survey, why do we need to create Observable for both `index and model`?

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,202 @@
...
+    private val loadMoreSurveysSingleUseCase: LoadMoreSurveysSingleUseCase,
+    private val updateSurveysTotalPagesSingleCompletableCase: UpdateSurveysTotalPagesSingleCompletableCase
 ) : BaseViewModel(), Inputs {
 
     private val _error = PublishSubject.create<Throwable>()
 
     private val _showLoading = BehaviorSubject.create<Boolean>()
 
-    private val _surveysPagerItemUiModels = BehaviorSubject.create<List<SurveysPagerItemUiModel>>()
+    private val _selectedSurveyData = BehaviorSubject.createDefault(Pair<Int, SurveyItemUiModel?>(DEFAULT_UNSELECTED_INDEX, null))
```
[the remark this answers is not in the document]
hoangnguyen92dn: When initialize, I don't think we need to `createDefault`, not sure about your idea about handling index with ui model. Please recheck my comment below 👇

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,202 @@
...
+
+    private var reachedLastSurveysPage = false
+
+    override fun nextIndex() {
+        val surveys = currentSurveyItemUiModels
+        val nextIndex = selectedSurveyIndexValue + 1
+        setDisplaySurveyItemWith(nextIndex)
+
+        // When we reach the last item, trigger to load more if needed
+        if (surveys.size - 1 == nextIndex && !reachedLastSurveysPage || surveys.size == nextIndex) {
```
hoangnguyen92dn, line 71: `surveys.size - 1 == nextIndex`
`surveys.size == nextIndex`

I am concerned about conditions above. we already have `reachedLastSurveysPage` to validate condition for loading more. Could we create separate function like `shouldLoadMore()` and invoke load more function if the function return `true`?

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,202 @@
...
+        }
+        if (shouldUpdatePageNumber) {
+            calculateInitialSurveysPageNumber(finalSurveys.size)
+        }
+        _surveyItemUiModels.onNext(finalSurveys)
+    }
+
+    private fun calculateInitialSurveysPageNumber(totalItems: Int) {
+        var pageNumber = totalItems / currentSurveysPageSize
+        if (pageNumber < 1) pageNumber = 1
```
hoangnguyen92dn, line 144: `pageNumber = 1`, could we use `DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER`?

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,202 @@
...
+            )
+            .bindForDisposable()
+    }
+
+    private fun mergeSurveys(newSurveys: List<SurveyItemUiModel>): List<SurveyItemUiModel> =
+        currentSurveyItemUiModels
+            .plus(newSurveys)
+            .distinctBy { survey -> survey.id }
+
+    private fun setDisplaySurveyItemWith(index: Int) {
```
hoangnguyen92dn, line 189: `setDisplaySurveyItemWith` a little bit confused for idea what we will do. I'd suggest to rename this method for clearer.

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,202 @@
...
+        if (shouldMerge) {
+            finalSurveys = mergeSurveys(surveys)
+        }
+        if (shouldUpdatePageNumber) {
+            calculateInitialSurveysPageNumber(finalSurveys.size)
+        }
+        _surveyItemUiModels.onNext(finalSurveys)
+    }
+
+    private fun calculateInitialSurveysPageNumber(totalItems: Int) {
```
hoangnguyen92dn, line 142: `calculateInitialSurveysPageNumber`, the name seems like for `initial` load but we are using it for every time surveys updated.

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,202 @@
...
+                GetInitialSurveysFlowableUseCase.Input(
+                    DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER,
+                    currentSurveysPageSize
+                )
+            )
+            .map { it.map { survey -> survey.toSurveyItemUiModel() } }
             .doFinally { _showLoading.onNext(false) }
             .subscribeBy(
                 onNext = {
                     if (it.isNotEmpty()) _showLoading.onNext(false)
```
hoangnguyen92dn, line 117: `_showLoading.onNext(false)`, we handle this from `doFinally` event the list is empty or not. Please recheck 🙏

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
manh-t: > Totally agree with you that we will need to handle no internet connection case. However, there is no requirement in the Android Certification for doing it so I am not doing an estimation for that task. I will implement it in a separate PR when the app is fully functional as an additional optimization. Sounds good to you?
Sounds good, please go ahead 😀 

> Previously I thought that your flow is trigger the network call to load per page -> save to database -> getting all data from database as per this sentence I think it is database in this case, we will fetch data from network, then inserting data to the database, and we can listen the change from database and update the UI accordingly.. If that is not your point then we can just ignore
yeah, I just share the knowledge that we can apply [Single source of truth](https://medium.com/@sina.rahimi/single-source-of-truth-with-mvvm-retrofit2-livedata-rxjava-and-room-in-repository-pattern-f5304f39175) as well since this is also a good alternative approach beside of yours 🤓 . When we use this approach, we just need to setup the listener to `database` once, and then whenever there is change in database (ex. when loading more data completely `.doOnSuccess { surveyDao.insertAll(it) }`), it automatically fills in new data into UI 😸  
Thanks for kindly answering my question 🙏

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
```diff
@@ -1,38 +1,36 @@
...
-        return Single.concat(
-            surveyDao.getAllSurveys(),
-            surveyService
+    ): Single<List<Survey>> {
+        return surveyService
             .getSurveysList(pageNumber, pageSize)
             .firstOrError()
-            .map { it.toSurveys() }
+            .doOnSuccess { appPreferences.surveysTotalPages = it.pages }
+            .map { it.surveys.toSurveys() }
```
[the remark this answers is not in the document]
manh-t: Agree, separation the concern is good, it will be easier to perform testing

app/src/main/java/co/nimblehq/data/error/AppError.kt
```diff
@@ -1,15 +1,16 @@
...
 open class AppError(
     cause: Throwable?,
     open val readableMessage: String? = null,
     @StringRes open val readableMessageRes: Int? = null
 ) : Throwable(cause) {
 
-    protected open val code: String?
-        get() = (cause as? JsonApiException)?.error?.code
+    protected open val code: Int?
+        get() = (cause as? JsonApiException)?.response?.code()
```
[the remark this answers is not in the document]
minhnimble: ah, right now our API response's error code returns nothing (empty string). As a result it can't be used and we have to use the HTTP code instead. 

I guess we need to ask our backend guys support it for this to work, right now it doesn't returns anything 😓

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,202 @@
...
+    private fun calculateInitialSurveysPageNumber(totalItems: Int) {
+        var pageNumber = totalItems / currentSurveysPageSize
+        if (pageNumber < 1) pageNumber = 1
+        currentSurveysPageNumber = pageNumber
+    }
+
+    private fun checkIfReachedLastSurveysPage() {
+        getSurveysTotalPagesSingleUseCase
+            .execute(Unit)
+            .map { it }
```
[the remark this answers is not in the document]
minhnimble: oh let me take a look, previously I was return an optional Int? and it requires the `.map` call, but I updated it to return Int instead but forget to remove here 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: we need `model` for displaying contents in the same observable. Is there any better approach? 🤔

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: Oh to clarify this combined condition, I would like to explain it as follow:
First condition `surveys.size - 1 == nextIndex && !reachedLastSurveysPage` means when users at last item, we will automatically trigger the load if `reachedLastSurveysPage` is false
Second condition `surveys.size == nextIndex` means when user is at the last item and the trigger a right swipe, we will show an error to help them know why there is nothing happening -> there is nothing more to load.

> Could we create separate function like shouldLoadMore() and invoke load more function if the function return true

Yeah sure, let me refactor

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: oh nice catch, let me update

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: nope, we have a flag check already, it is only true when we are triggering the initial load 😊
Screen Shot 2020-10-22 at 10 39 02

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: Not sure if you looked into the use case yet but this is a Flowable event, I want to display the `database` data and stop the loading as soon as we have it, otherwise we will dismiss it with the finally call when the API returns so we have to put a check to dismiss earlier here

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: `updateCurrentSurveyAt` sounds good to you? 🤔

app/src/main/java/co/nimblehq/usecase/survey/UpdateSurveysTotalPagesSingleCompletableCase.kt
```diff
@@ -0,0 +1,24 @@
+package co.nimblehq.usecase.survey
+
+import co.nimblehq.data.error.Ignored
+import co.nimblehq.data.lib.schedulers.RxSchedulerProvider
+import co.nimblehq.data.storage.AppPreferences
+import co.nimblehq.usecase.base.CompletableUseCase
+import io.reactivex.Completable
+import javax.inject.Inject
+
+class UpdateSurveysTotalPagesSingleCompletableCase @Inject constructor(
```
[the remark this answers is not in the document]
minhnimble: oh my stupid mistake, let me update 👍

app/src/main/java/co/nimblehq/usecase/survey/GetInitialSurveysFlowableUseCase.kt
```diff
@@ -24,9 +27,9 @@ class GetSurveysListFlowableUseCase @Inject constructor(
 
     override fun create(input: Input): Flowable<List<Survey>> {
         return with(input) {
-            surveyRepository.getSurveysList(
-                pageNumber,
-                pageSize
+            Single.concat(
+                surveyDao.getAllSurveys(),
```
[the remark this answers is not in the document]
minhnimble: I have implemented that logic as you mentioned that we will try to get data from database and display at the same time trigger the API call to update, not sure how to produce your case... Can you give me a details flow to reproduce that?

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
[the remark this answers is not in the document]
minhnimble: This case we need default values because in the case the API call fails and all other checks can still work when there is no data available and user triggers actions. Not sure why you are against `createDefault`, some situations we will need it like this case.

> It would break max character per line rule. Please reformat 🙏

Yeah sure, I will update 👍

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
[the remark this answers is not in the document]
minhnimble: okay I can move this logic to use case for clear separation in logic. This method is inspired by The1 Android project 🙏

app/src/main/java/co/nimblehq/data/error/AppError.kt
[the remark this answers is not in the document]
hoangnguyen92dn: Yeah. Got your idea, because I don't see we use http code for any purpose so that's make me confused. We could ignore it for now.

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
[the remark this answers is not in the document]
hoangnguyen92dn: Actually, because you are using 2 UseCase for fetching surveys:

`GetInitialSurveysFlowableUseCase`: For database and remote data
`LoadMoreSurveysSingleUseCase`: For remote data

Why don't we use `GetInitialSurveysFlowableUseCase` for loading from database only? Then we only have 1 UseCase for the list of surveys from remote -> You could execute UseCase for updating SharedPreferences from ViewModel. That's all. Does it make sense?

app/src/main/java/co/nimblehq/usecase/survey/GetInitialSurveysFlowableUseCase.kt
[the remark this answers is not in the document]
hoangnguyen92dn: Sorry for the unclear procedure steps

SVID_20201022_143401_1

I think the issue as I mentioned related to connectivity, I just tried to turn off connection to make sure the app always display data from database first before displaying new data from remote.

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
hoangnguyen92dn: > Maybe I can update to use only selectedIndex observable and and get the surveyItemUiModel from the array instead, sounds good to you?

Yeah. Makes sense.

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
hoangnguyen92dn: Yeah. Sounds good.

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
hoangnguyen92dn: I mean the `initial` from method name makes me confused 😅

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
manh-t: hmm yeah, I got your point, but the naming makes me confused a bit when applying here. No worries, you can keep it as it is.
We could have another solution that we create a use case for getting surveys from database, and other use case for getting surveys from network. Therefore, it would be cleaner imo, it's up to you 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
hoangnguyen92dn: Hmmm. Got your point. Because you are combing 2 single stream in the `UseCase`. As I suggested above, we could create 2 separate UseCase and use `map/flatMap` to execute `database` and `remote` UseCases in stream.

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: actually, I agree with you on the confusing part, I just updated the usecase name for clear separation. You can check again 😁

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: oh I understand your point now, I will remove the word `initial` 👍

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: yeah let me do that. This way we can have clean separation and not having to combine the if check into this 👍

app/src/main/java/co/nimblehq/data/repository/SurveyRepository.kt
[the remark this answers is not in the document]
minhnimble: okay your explain is clearer now, I think doing your suggestion is fine as well. I can update to it.

app/src/main/java/co/nimblehq/usecase/survey/GetInitialSurveysFlowableUseCase.kt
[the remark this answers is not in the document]
minhnimble: I don't have this issue on my end using your steps. Can you try and checkout the latest code if this branch and check again? Thanks 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
```diff
@@ -31,43 +33,76 @@ class SurveysFragment: BaseFragment(), BaseFragmentCallbacks {
...
+        btSurveysItemNext.subscribeOnClick {
+            // TODO: Handle navigate to survey details screen
+        }.bindForDisposable()
+
+        clSurveys.setOnTouchListener(object : OnSwipeTouchListener(requireContext()) {
+
+            override fun onSwipeBottom() {
+                super.onSwipeBottom()
+                viewModel.refreshSurveysList()
+            }
```
manh-t, line 53: hmm, I think you should handle pull to refresh in other way. We often use SwipeRefreshLayout for being better UI/UX 🤔

minhnimble: hmm, thanks for your suggestion! Let me investigate and try it

minhnimble: it is done here 034f31f, you can take a look now, thanks for your suggestion 🙏 really clean 👍😁

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,197 @@
...
+            calculateSurveysPageNumber(finalSurveys.size)
+        }
+        _surveyItemUiModels.onNext(finalSurveys)
+    }
+
+    private fun calculateSurveysPageNumber(totalItems: Int) {
+        var pageNumber = totalItems / currentSurveysPageSize
+        if (pageNumber < DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER) pageNumber = DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER
+        currentSurveysPageNumber = pageNumber
+    }
```
hoangnguyen92dn, line 158: Just curious about this, we already know the page will be loaded and API also return `page` so why do we need to do computation here? 🤔

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,197 @@
...
+    private var currentSurveysPageSize = DEFAULT_SURVEYS_PAGE_SIZE
+
+    private var reachedLastSurveysPage = false
+
+    override fun nextIndex() {
+        val nextIndex = selectedSurveyIndexValue + 1
+        updateSurveyIndex(nextIndex)
+
+        if (shouldLoadMoreSurveys(nextIndex)) { // When we reach the last item, trigger to load more if needed
+            checkAndLoadMoreSurveysIfNeeded()
```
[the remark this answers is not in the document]
minhnimble: oh I understand your point, let me update this logic a bit 👍

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,197 @@
...
     }
+
+    fun getSelectedSurveyUiModel(): SurveyItemUiModel? {
+        val models = currentSurveyItemUiModels
+        val selectedIndex = selectedSurveyIndexValue
+        return if (models.isValidIndex(selectedIndex)) {
+            models[selectedIndex]
+        } else {
+            null
+        }
```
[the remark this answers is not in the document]
minhnimble: oh we have it? I am totally not aware of it, nice suggestion thanks 👍

app/src/main/java/co/nimblehq/usecase/survey/DeleteLocalSurveysCompletableUseCase.kt
```diff
@@ -0,0 +1,29 @@
+package co.nimblehq.usecase.survey
+
+import co.nimblehq.data.error.SurveyError.DeleteLocalSurveysError
+import co.nimblehq.data.lib.schedulers.RxSchedulerProvider
+import co.nimblehq.data.storage.dao.SurveyDao
+import co.nimblehq.usecase.base.CompletableUseCase
+import io.reactivex.Completable
+import javax.inject.Inject
+
+class DeleteLocalSurveysCompletableUseCase @Inject constructor(
```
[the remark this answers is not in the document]
minhnimble: hmm, per the new logic, I think it is better to clear all data and only show the first page since we are getting only the first page

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,197 @@
...
+    fun refreshSurveysList() {
+        loadSurveysSingleUseCase
+            .execute(LoadSurveysSingleUseCase.Input(DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER, currentSurveysPageSize))
+            .doOnSubscribe { _showRefreshing.onNext(true) }
+            .map { it.map { survey -> survey.toSurveyItemUiModel() } }
+            .doOnSuccess {
+                bindOnSuccessLoadSurveys(it, shouldMerge = false)
+                updateSurveyIndex(0)
+            }
+            .flatMapCompletable { deleteLocalSurveysCompletableUseCase.execute(DeleteLocalSurveysCompletableUseCase.Input(it.map { model -> model.id })) }
```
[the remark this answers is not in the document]
minhnimble: I think this is a good point since we only do refresh for first page, let me update 👍

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,197 @@
...
+
+    fun checkAndRefreshInitialSurveys() {
+        getLocalSurveysSingleUseCase
+            .execute(Unit)
+            .map { it.map { survey -> survey.toSurveyItemUiModel() } }
+            .doOnSuccess {
+                bindOnSuccessLoadSurveys(it, shouldMerge = false)
+                if (it.isNotEmpty()) {
+                    _showLoading.onNext(false)
+                    updateSurveyIndex(0)
```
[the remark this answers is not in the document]
minhnimble: hmm I just want to dismiss the loading as soon as we have something to show, but in our case I guess we can update to your suggestion

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,197 @@
...
+                }
+            }
+            .flatMap { loadSurveysSingleUseCase.execute(LoadSurveysSingleUseCase.Input(DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER, currentSurveysPageSize)) }
+            .map { it.map { survey -> survey.toSurveyItemUiModel() } }
+            .doOnSuccess {
+                bindOnSuccessLoadSurveys(it)
+                if (selectedSurveyIndexValue == DEFAULT_UNSELECTED_INDEX) updateSurveyIndex(0)
```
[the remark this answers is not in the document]
minhnimble: nope, if we have data in `database`, `selectedSurveyIndexValue` will not be equal to DEFAULT_UNSELECTED_INDEX and thus this condition will not be triggered. 

I choose `-1` instead of `0` for `default selected index` because `0` is a valid index in an array. `-1` is for representing the case initially we have no data.

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,197 @@
...
+    private val currentSurveyItemUiModels: List<SurveyItemUiModel>
+        get() = _surveyItemUiModels.value.orEmpty()
+
+    private var currentSurveysPageNumber = DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER
 
-    fun checkAndLoadSurveysListIfNeeded(pageNumber: Int = DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER, pageSize: Int = DEFAULT_SURVEYS_PAGE_SIZE) {
-        getSurveysListFlowableUseCase
-            .execute(GetSurveysListFlowableUseCase.Input(pageNumber, pageSize))
-            .map { it.map { survey -> survey.toSurveysPagerItemUiModel() } }
+    private var currentSurveysPageSize = DEFAULT_SURVEYS_PAGE_SIZE
```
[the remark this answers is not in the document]
minhnimble: oh this is for having a place you can update the page size to your need for easier testing with different pageSize to see how the logic works. This is for future usage 😊

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: hmm I am not parsing that value at the moment, I think I can parse it actually, nice thought 👍

manh-t (approved):

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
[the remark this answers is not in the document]
hoangnguyen92dn: Sorry I was a bit busy so just back to this PR 😅  

I got your idea. Thanks for correct me here. Just wondering, why don't we clear all and insert the refreshed page (actually always the first page from API) as simplest and for better performance 🤔  Any specific reason here?

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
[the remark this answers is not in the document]
hoangnguyen92dn: 👌

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,183 @@
...
+        if (shouldMerge) {
+            finalSurveys = mergeSurveys(surveys)
+        }
+        _surveyItemUiModels.onNext(finalSurveys)
+        if (shouldResetToFirstIndex) {
+            updateSurveyIndex(0)
+        }
+    }
+
+    private fun getLocalPagingAttributes(): Single<Int> {
```
hoangnguyen92dn, line 144: This function returns `Single<Int>` but we are not using return value from this `Single`? 🤔

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,183 @@
...
+        return getSurveysCurrentPageSingleUseCase.execute(Unit)
+            .doOnSuccess { currentSurveysPageNumber = it.coerceAtLeast(DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER) }
```
hoangnguyen92dn, line 146: Ideally, we always know current page and use it for API request, imo we don't need to save the `page` from API response as we used to for sending request. Does it make sense?

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,183 @@
...
+            .flatMap { getSurveysTotalPagesSingleUseCase.execute(Unit) }
+            .doOnSuccess { reachedLastSurveysPage = it in DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER..currentSurveysPageNumber }
```
hoangnguyen92dn, line 148: Any reason why don't we just compare if `it` equals to return value from `getSurveysTotalPagesSingleUseCase`?

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,183 @@
...
+
+    fun checkAndRefreshInitialSurveys() {
+        getLocalSurveysSingleUseCase
+            .execute(Unit)
+            .map { it.toSurveyItemUiModels() }
+            .doOnSuccess { bindOnSuccessLoadSurveys(it) }
+            .flatMap { getLocalPagingAttributes() }
+            .flatMap { loadSurveysSingleUseCase.execute(LoadSurveysSingleUseCase.Input(DEFAULT_INITIAL_SURVEYS_PAGE_NUMBER, currentSurveysPageSize)) }
+            .map { it.toSurveyItemUiModels() }
+            .doOnSuccess { bindOnSuccessLoadSurveys(it, shouldMerge = false, shouldResetToFirstIndex = selectedSurveyIndexValue == DEFAULT_UNSELECTED_INDEX) }
```
hoangnguyen92dn, line 110: `shouldResetToFirstIndex = selectedSurveyIndexValue == DEFAULT_UNSELECTED_INDEX)`

As far as I understand, the condition above always `true`, we are emit the list from local database but `loading` still showing so user can not change the `selectedSurveyIndexValue` from UI by swiping left/right so it always as default. Please correct me if I am wrong.

app/src/main/java/co/nimblehq/data/storage/AppPreferences.kt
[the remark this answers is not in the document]
hoangnguyen92dn: I'd like to recall this discussion because the new logic updated from `checkAndRefreshInitialSurveys` in `SurveysViewModel`

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,183 @@
...
+    }
+
+    private fun shouldLoadMoreSurveys(newIndex: Int): Boolean {
+        return currentSurveyItemUiModels.indices.lastOrNull()?.let { lastIndex ->
+            // Notify user when there is nothing more to load
+            if (newIndex == lastIndex + 1 && reachedLastSurveysPage) {
+                _error.onNext(SurveyError.NoMoreSurveysError(null))
+            }
+            newIndex in (lastIndex..lastIndex + 1) && !reachedLastSurveysPage
+        } ?: run { false }
```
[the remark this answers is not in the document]
minhnimble: nice catch, let me update

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,183 @@
...
+        _selectedSurveyIndex.onNext(newIndex)
+    }
+
+    private fun shouldLoadMoreSurveys(newIndex: Int): Boolean {
+        return currentSurveyItemUiModels.indices.lastOrNull()?.let { lastIndex ->
+            // Notify user when there is nothing more to load
+            if (newIndex == lastIndex + 1 && reachedLastSurveysPage) {
+                _error.onNext(SurveyError.NoMoreSurveysError(null))
+            }
+            newIndex in (lastIndex..lastIndex + 1) && !reachedLastSurveysPage
```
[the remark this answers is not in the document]
minhnimble: yeah that is intentional logic because when user swipes to the last item, if we don't use `in`, user will have to swipe one more time to trigger the load, but I want to trigger the load next page as soon as we reach the last item, not having to trigger another swipe while on last time.

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,183 @@
...
+    private fun mergeSurveys(newSurveys: List<SurveyItemUiModel>): List<SurveyItemUiModel> =
+        currentSurveyItemUiModels.plus(newSurveys).distinctBy { it.id }
+
+    private fun updateSurveyIndex(newIndex: Int) {
+        if (currentSurveyItemUiModels.isNotValidIndex(newIndex)) return
+        _selectedSurveyIndex.onNext(newIndex)
+    }
+
+    private fun shouldLoadMoreSurveys(newIndex: Int): Boolean {
+        return currentSurveyItemUiModels.indices.lastOrNull()?.let { lastIndex ->
```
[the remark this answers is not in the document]
minhnimble: nice catch, let me update 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -1,51 +1,183 @@
...
             )
             .bindForDisposable()
     }
+
+    private fun mergeSurveys(newSurveys: List<SurveyItemUiModel>): List<SurveyItemUiModel> =
+        currentSurveyItemUiModels.plus(newSurveys).distinctBy { it.id }
+
+    private fun updateSurveyIndex(newIndex: Int) {
+        if (currentSurveyItemUiModels.isNotValidIndex(newIndex)) return
+        _selectedSurveyIndex.onNext(newIndex)
```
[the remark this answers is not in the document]
minhnimble: I think your suggestion is ok, will update 👍

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: yeah we use it here:
Screen Shot 2020-11-02 at 14 02 51
and here
Screen Shot 2020-11-02 at 14 02 57
Outside we just need to chain and process the next step. 😊

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: This would help will data integrity and single source of truth, as discussed with Manh that we can use the current page from the response

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: I don't understand your point, `it` is the value `from getSurveysTotalPagesSingleUseCase` 🤔. you mean comparing `it` with `currentSurveysPageNumber` ?

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
minhnimble: that is not true, this check is **not** for checking `user changes UI by swiping left/right`. It is for checking the first load from database, if we have data from database, we already default the `selectedItem` to `firstIndex` already, and thus this check here will return `false`. This check is only true if the first load from database returns empty content.

minhnimble: but I think it is true that we can definitely remove the flag here since setting the first item index to 0 again is acceptable because like you said, user cannot change `selectedSurveyIndexValue` from UI by swiping left/right

hoangnguyen92dn (approved):

PR #19, opened 2020-10-24
[TASK-8] [Logic] As a user, I can select to take a survey or go back

minhnimble:
https://github.com/minhnimble/android-surveys-list/projects/1#card-46083260

## What happened

When users are looking at the surveys list, they will be able to select and take a displayed survey or just look at that survey details and go back to the surveys list.

This PR is for creating the logic for loading survey details with a list of questions that when user clicks `Start Survey` button, the app will have data to show the survey's questions list.
 
## Insight

- [x] Connect to survey details API to get latest survey's questions list
 
## Proof Of Work

Request survey details data successfully:
Screen Shot 2020-10-24 at 16 15 18


app/src/main/java/co/nimblehq/data/model/Answer.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import androidx.room.ColumnInfo
+import androidx.room.PrimaryKey
+import co.nimblehq.data.api.response.survey.AnswerResponse
+
+data class Answer(
+    @PrimaryKey
+    @ColumnInfo(name = "id") val id: String = "",
+    @ColumnInfo(name = "text") val text: String = "",
+    @ColumnInfo(name = "display_order") val displayOrder: Int = -1
```
manh-t, line 11: I see that you currently don't use `Answer` and `Question` in database. Are you going to add them into database in the future? 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
```diff
@@ -2,13 +2,46 @@ package co.nimblehq.ui.screen.main.surveydetails
...
+    private val loadSurveyDetailsSingleUseCase: LoadSurveyDetailsSingleUseCase
 ) : BaseViewModel(), Inputs {
 
+    private val _getSurveyDetailsError = PublishSubject.create<Throwable>()
+
+    private val _showLoading = BehaviorSubject.create<Boolean>()
+
+    private val _questionItemPagerUiModels = BehaviorSubject.create<List<QuestionItemPagerUiModel>>()
+
+    val getSurveyDetailsError: Observable<Throwable>
```
manh-t, line 24: I think `error` should not go with `get` 🤔 , or we can name it as `showError` and handle all errors here 🙏

data/src/main/java/co/nimblehq/data/api/parser/JsonReaderExtension.kt
```diff
@@ -6,6 +6,27 @@ fun JsonReader.skipNameAndValue() {
...
+        ""
+    } else {
+        nextString()
+    }
+}
+
+fun JsonReader.nextIntOrEmpty(): Int {
+    return if (isNextNull()) {
+        nextNull<Int>()
+        0
```
manh-t, line 24: I'm not sure if we should handle `null` int value as `0`. If the data comes from API, I think we could mark them as nullable, including `nextStringOrEmpty`. How do you think?

data/src/main/java/co/nimblehq/data/api/parser/SurveyResponseParser.kt
```diff
@@ -1,6 +1,6 @@
 package co.nimblehq.data.api.parser
```
manh-t, line 1: The parser seems to be complicated 😅 
Can I ask that are you trying to parse json manually? 🙏   If so, I'm curious that how can you deal with this if the application is scalable? 😀

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
```diff
@@ -1,16 +1,76 @@
...
+    @Json(name = "questions") var questions: List<QuestionResponse> = listOf()
 )
 
 data class SurveysResponse(
     @Json(name = "page") var page: Int = 0,
     @Json(name = "pages") var pages: Int = 0,
     @Json(name = "surveys") var surveys: List<SurveyResponse> = listOf(),
 )
+
+data class QuestionResponse(
```
manh-t, line 20: IMO, we should create separate file for `QuestionReponse`. How do you think?

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
```diff
@@ -1,16 +1,76 @@
...
+    @Json(name = "id") var id: String = "",
+    @Json(name = "text") var text: String = "",
+    @Json(name = "display_order") var displayOrder: Int = -1,
+    @Json(name = "display_type") var displayType: QuestionDisplayType = QuestionDisplayType.DEFAULT,
+    @Json(name = "pick") var pick: QuestionPickValue = QuestionPickValue.NONE,
+    @Json(name = "answers") var answers: List<AnswerResponse> = listOf()
```
manh-t, line 26: should the values be nullable when getting data from API as I mentioned above? ☝️  We don't know that when will the API will return error case (eg. missing `display_order`), or data from API is the same as our default value 🤔

manh-t:
Sorry for the late review @minhnimble 🙇  I added some questions into the PR, please help me take a look at them 😁

app/src/main/java/co/nimblehq/data/model/Answer.kt
minhnimble: nice question 🙏. About this part, when reading requirement, there is no mention about having to store Questions and Answers to database. Instead, the requirements ask us to implement questions list based on the API response:
`On the survey details page, the list of questions must be implemented based on the API response`

From my understanding, this would mean we have to get the latest survey questions list every time we start the survey, making sure we have latest questions data. As a result, I don't think we need to store questions and answers to cache 😊

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
minhnimble: oh okay, in this screen there is only one type of error, which is `GetSurveyDetailsError`, that is why I name it like this. I guess I will go with your suggestion for consistency 👍

data/src/main/java/co/nimblehq/data/api/parser/JsonReaderExtension.kt
minhnimble: oh, yeah I think it works for me too, let me update 👍

data/src/main/java/co/nimblehq/data/api/parser/SurveyResponseParser.kt
minhnimble: previously I use automatic parsing but Trung asks me to flatten the response to only the objects we need, that is why I do this. My experiences with JSON parsing is not very much so this is the best I can do for now 😅 If you have any good suggestions for this case I would love learn more 💪

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
minhnimble: I think that is okay, will do it for clean separation

minhnimble:
Hi @manh-t, no worries, thank you for taking your time to review my branch 😍 taking a look now 🙏

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
minhnimble: I think providing default values is a much better approach, because eventually when you use the `response` in `ViewModel`, and then when the value is `null, we will also have to default it to something for the display to work. Maybe I can send the default assignments from `Response` layer to `Model` layer instead, and will update `Response` layer will nullable attributes for correctly reflecting API response data, what do you think? 😊

data/src/main/java/co/nimblehq/data/api/parser/JsonReaderExtension.kt
minhnimble: same [comment](https://github.com/minhnimble/android-surveys-list/pull/19#discussion_r514981427) here  🙏

app/src/main/java/co/nimblehq/data/model/Answer.kt
manh-t: > As a result, I don't think we need to store questions and answers to cache

yes, I think so. We can display the error when there is no connectivity in case fetching the question. 
So shall we remove the annotations? (e.g `ColumnInfo`, `PrimaryKey`)😊

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
manh-t: > send the default assignments from Response layer to Model layer instead, and will update Response layer will nullable attributes for correctly reflecting API response data like your suggestion

@minhnimble yeah, I totally agree 🚀  Please go ahead

app/src/main/java/co/nimblehq/data/model/Answer.kt
minhnimble: yeah sure, will update to remove, thanks 👍

data/src/main/java/co/nimblehq/data/api/parser/SurveyResponseParser.kt
manh-t: From my understanding, we can create some base responses based on API response structure 🤔 
The base data always return
```json
// success:
{"data":{}}
// or 
{"data":[]}

//failed
{"error":{}}
```
So that we can create `BaseResponse` like this:
```kotlin
data class BaseResponse(
    @SerializedName("data") val data: Any?, // because the we are not sure that the data is object or list of object
    @SerializedName("error") val error: BaseErrorResponse?, // you can create the error response yourself easily
    @SerializedName("included") val included: List<??> // add if needed
)
```
I believe that the `data` object is consistency throughout app, it is:
```kotlin
data class DataResponse<T>(
    @SerializedName("id") val id: Any?, // I see that it's int/String sometimes in API
    @SerializedName("type") val type: String?,
    @SerializedName("attributes") val attribute: T   // T is our main thing
)
```

This nearly reach the target 😛 
We can cast the `data` in `BaseResponse` to `DataResponse<T>` or `List<DataResponse<T>>` according to the need.
This is my thought. Sorry if I missed something or my approach is wrong and hard to understand to you 😭  🙏

app/src/main/java/co/nimblehq/data/model/Question.kt
```diff
@@ -0,0 +1,32 @@
...
+import co.nimblehq.data.api.response.survey.QuestionDisplayType
+import co.nimblehq.data.api.response.survey.QuestionPickValue
+import co.nimblehq.data.api.response.survey.QuestionResponse
+import co.nimblehq.data.storage.dao.converter.QuestionDisplayTypeConverter
+import co.nimblehq.data.storage.dao.converter.QuestionPickValueConverter
+
+@TypeConverters(
+    QuestionDisplayTypeConverter::class,
+    QuestionPickValueConverter::class
+)
```
manh-t, line 13: I think you missed this. We don't store Question or Answer to database anymore, so we could get rid of `@TypeConverters` here

data/src/main/java/co/nimblehq/data/api/parser/SurveyResponseParser.kt
minhnimble: Hi @manh-t , I initially did this and understand your point, then Trung askes me to implement a custom Parser so that in the end it just returns the exact data we need as per comments from here `https://github.com/minhnimble/android-surveys-list/pull/12#discussion_r503842414` and here `https://github.com/minhnimble/android-surveys-list/pull/12#discussion_r504374424` 😅

app/src/main/java/co/nimblehq/data/model/Question.kt
minhnimble: yeah you are right, let me get rid of them as well, sorry previously you just mention about `@ColumnInfo` `PrimaryKey` so I forget 🙏

data/src/main/java/co/nimblehq/data/api/parser/SurveyResponseParser.kt
manh-t: so 🆗  🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveyItemUiModel.kt
```diff
@@ -19,4 +20,6 @@ fun Survey.toSurveyItemUiModel() = SurveyItemUiModel(
     highResImageUrl
 )
 
+fun Survey.toQuestionItemPagerUiModels() = questions.toQuestionItemPagerUiModels()
```
manh-t, line 23: I think this function should be in `QuestionItemPagerUiModel.kt` 🙏

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
```diff
@@ -6,11 +6,12 @@ data class SurveyResponse(
...
+    @Json(name = "cover_image_url") var coverImageUrl: String = "",
+    @Json(name = "questions") var questions: List<QuestionResponse> = listOf()
 )
 
 data class SurveysResponse(
-    @Json(name = "page") var page: Int = 0,
-    @Json(name = "pages") var pages: Int = 0,
+    @Json(name = "page") var page: Int = 1,
+    @Json(name = "pages") var pages: Int = 1,
     @Json(name = "surveys") var surveys: List<SurveyResponse> = listOf(),
```
manh-t, line 16: should we apply nullable objects for `SurveyResponse` and `SurveysResponse` as well? 😅

app/src/main/java/co/nimblehq/data/model/Question.kt
```diff
@@ -0,0 +1,25 @@
...
+    val answers: List<Answer> = listOf()
+)
+
+fun QuestionResponse.toQuestion() = Question(
+    id = id,
+    text = text ?: "",
+    displayOrder = displayOrder ?: -1,
+    displayType = displayType ?: QuestionDisplayType.DEFAULT,
+    pick = pick ?: QuestionPickValue.NONE,
+    answers = (answers ?: listOf()).toAnswers()
```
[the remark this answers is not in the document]
minhnimble: okay, will update 👍

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
minhnimble: yeah that should do for consistent 😊

data/src/main/java/co/nimblehq/data/api/parser/SurveyResponseParser.kt
```diff
@@ -10,25 +10,106 @@ class SurveyResponseParser: JsonAdapter<SurveyResponse>() {
...
                 when (reader.nextName()) {
                     "id" -> surveyResponse.id = reader.nextString()
-                    "attributes" -> {
-                        reader.readObject {
-                            when (reader.nextName()) {
-                                "title" -> surveyResponse.title = reader.nextString()
-                                "description" -> surveyResponse.description = reader.nextString()
-                                "cover_image_url" -> surveyResponse.coverImageUrl = reader.nextString()
-                                else -> reader.skipValue()
+                    "included" -> {
```
hoangnguyen92dn, line 20: I got your idea about using `JsonAdapter` for flatten objects from `json api` with these attributes such as `included`, 'relationships',... (FYI: https://jsonapi.org/) So as I can see, we already added `banana` library as dependencies, just curious why don't we use `banana` for parsing `json api` like the way The1 implemented rather than doing manually here. (Refer: `ResourceAdapterProvider` from The1 Android project and usage of `JsonApi` annotation)

Thanks @luongvo for explanation about Moshi and Banana to me 🙏

hoangnguyen92dn (changes requested):

data/src/main/java/co/nimblehq/data/api/parser/SurveyResponseParser.kt
minhnimble: Yeah, originally I spent 2 days to investigate how we can use a Parser to flatten objects with` https://jsonapi.org/` `Moshi` but unfortunately, I could not make the parser works correct to `jsonapi` style previously. As a result, I have to do it this way. 

The issue I have previously is that, The1 response object is already have all params grouped inside `attributes`, but here in the test project api, the `id` is outside of `attributes` object so how can you flatten `id` from outside `attributes` and other `attributes` keys into one single flatten object?

app/src/main/java/co/nimblehq/data/model/Answer.kt
```diff
@@ -0,0 +1,17 @@
...
+
+data class Answer(
+    val id: String = "",
+    val text: String = "",
+    val displayOrder: Int = -1
+)
+
+fun AnswerResponse.toAnswer() = Answer(
+    id = id,
+    text = text ?: "",
```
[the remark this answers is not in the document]
minhnimble: got it

app/src/main/java/co/nimblehq/data/model/Question.kt
```diff
@@ -0,0 +1,25 @@
...
+    var text: String = "",
+    var displayOrder: Int = -1,
+    var displayType: QuestionDisplayType = QuestionDisplayType.DEFAULT,
+    var pick: QuestionPickValue = QuestionPickValue.NONE,
+    val answers: List<Answer> = listOf()
+)
+
+fun QuestionResponse.toQuestion() = Question(
+    id = id,
+    text = text ?: "",
```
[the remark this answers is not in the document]
minhnimble: got it

app/src/main/java/co/nimblehq/usecase/survey/LoadSurveyDetailsSingleUseCase.kt
```diff
@@ -0,0 +1,31 @@
...
+    private val surveyRepository: SurveyRepository
+) : SingleUseCase<LoadSurveyDetailsSingleUseCase.Input, Survey>(
+    rxSchedulerProvider.io(),
+    rxSchedulerProvider.main(),
+    ::GetSurveyDetailsError
+) {
+
+    data class Input(
+        val surveyId: String
+    )
```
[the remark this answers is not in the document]
minhnimble: oh nice catch, forget about this 👍

app/src/main/java/co/nimblehq/data/model/Answer.kt
```diff
@@ -0,0 +1,17 @@
+package co.nimblehq.data.model
+
+import co.nimblehq.data.api.response.survey.AnswerResponse
+
+data class Answer(
+    val id: String = "",
+    val text: String = "",
+    val displayOrder: Int = -1
```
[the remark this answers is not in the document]
minhnimble: I think you are conflicting with Manh's comment, we should make `Response` object nullable and provide default values in model/uimodel layer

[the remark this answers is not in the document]
hoangnguyen92dn: @manh-t I am not sure why you suggested to update to define default value from model/uimodel, we should define default value from response layer because the response from api could changed sometimes so adding default value for handle case response missing some expected elements. Can you please explain your idea?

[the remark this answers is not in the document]
hoangnguyen92dn: @manh-t I am not sure why you suggested to update to define default value from model/uimodel, we should define default value from response layer because the response from api could changed sometimes so adding default value for handle case response missing some expected elements. Can you please explain your idea?

data/src/main/java/co/nimblehq/data/api/parser/SurveyResponseParser.kt
hoangnguyen92dn: Yeah. I am worry about new coming feature with new api response so you still need to parse response manually, it would take unnecessary effort on that.

minhnimble: yeah totally understand your point, will check this

app/src/main/java/co/nimblehq/data/model/Answer.kt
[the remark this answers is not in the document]
manh-t: @hoangnguyen92dn We can define default value anywhere, but I think it's better if @minhnimble can know our formatting from the beginning (nullable at Response layer, then define value in Model layer), we can handle case response missing some expected elements in Model layer as well 😁

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/QuestionItemPagerUiModel.kt
```diff
@@ -0,0 +1,27 @@
...
+import co.nimblehq.data.api.response.survey.QuestionPickValue
+import co.nimblehq.data.model.Question
+import co.nimblehq.data.model.Survey
+
+data class QuestionItemPagerUiModel(
+    val id: String = "",
+    val text: String = "",
+    val displayType: QuestionDisplayType = QuestionDisplayType.DEFAULT,
+    val pick: QuestionPickValue = QuestionPickValue.NONE,
+    val answers: List<AnswerItemUiModel> = listOf()
```
manh-t, line 13: I think you don't need to specify the default values for UiModel since we get them from Model layer 🤔

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
```diff
@@ -2,13 +2,45 @@ package co.nimblehq.ui.screen.main.surveydetails
...
+import io.reactivex.subjects.PublishSubject
 
 interface Inputs { }
 
 class SurveyDetailsViewModel @ViewModelInject constructor(
-    // TODO: Add needed use cases here
+    private val loadSurveyDetailsSingleUseCase: LoadSurveyDetailsSingleUseCase
 ) : BaseViewModel(), Inputs {
 
+    private val _getSurveyDetailsError = PublishSubject.create<Throwable>()
```
manh-t, line 17: Should it be `_showError`?

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/QuestionItemPagerUiModel.kt
minhnimble: sure, let me update 👍

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
minhnimble: nice catch, I missed this place, let me update 👌

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/QuestionItemPagerUiModel.kt
minhnimble: okay that makes sense, let me update

data/src/main/java/co/nimblehq/data/api/parser/SurveyResponseParser.kt
minhnimble: Implemented the Moshi parser with Banana for JsonApi

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
```diff
@@ -1,16 +1,66 @@
...
+                        allAnswers!!.firstOrNull { it.id == answer.id }?.let { fullAnswer ->
+                            answers[answerIndex] = fullAnswer
+                        }
+                    }
+                    fullQuestion.fullAnswers = answers
+                    questions[questionIndex] = fullQuestion
+                }
+            }
+        }
+        return if (hasQuestion) questions else null
```
hoangnguyen92dn, line 37: We already applied `Moshi` and `Banana` so I believe we could get rid of this.

data/src/main/java/co/nimblehq/data/api/providers/ResourceAdapterProvider.kt
```diff
@@ -0,0 +1,18 @@
...
+import moe.banana.jsonapi2.ResourceAdapterFactory
+
+class ResourceAdapterProvider {
+    val resourceFactory: ResourceAdapterFactory
+        get() = ResourceAdapterFactory
+            .builder()
+            .add(
+                SurveyResponse::class.java,
+                QuestionResponse::class.java,
+                AnswerResponse::class.java
```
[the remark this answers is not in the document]
minhnimble: yep we don't need it with the current implementation, I just leave it there for backing up so I can revert to use the old parsing way. Will remove

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysViewModel.kt
```diff
@@ -94,7 +94,9 @@ class SurveysViewModel @ViewModelInject constructor(
             }
             .doFinally { _showRefreshing.onNext(false) }
             .subscribeBy(
-                onError = { _error.onNext(it) }
+                onError = {
+                    _error.onNext(it)
+                }
```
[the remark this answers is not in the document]
minhnimble: I like bracket style more, but this should be on 1 line, thanks for spotting 👍

data/src/main/java/co/nimblehq/data/api/response/survey/QuestionResponse.kt
```diff
@@ -0,0 +1,20 @@
...
+@JsonApi(type = "question")
+data class QuestionResponse(
+    @field: Json(name = "text") var text: String? = null,
+    @field: Json(name = "display_order") var displayOrder: Int? = null,
+    @field: Json(name = "display_type") var displayType: String? = null,
+    @field: Json(name = "pick") var pick: String? = null,
+    private var answers: HasMany<AnswerResponse>? = null,
+    var fullAnswers: List<AnswerResponse>? = null
+): Resource() {
+    fun getAnswerResponses(): List<AnswerResponse>? {
```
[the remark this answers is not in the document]
minhnimble: will update 👍

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
minhnimble: previously I have null issues when just do `questions?.get(document)` so I have to do this. Let me check

minhnimble: Added a better check for this case, you can take a look now 🙏

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
```diff
@@ -1,16 +1,34 @@
...
+    @field: Json(name = "pages") var pages: Int? = null
 )
+
+val List<SurveyResponse>.meta: SurveysMetaResponse?
+    get() = when {
+        isEmpty() -> SurveysMetaResponse()
+        else -> first().document.meta.get<SurveysMetaResponse>(
+            Operators.moshi.adapter(SurveysMetaResponse::class.java)
+        ) as? SurveysMetaResponse
+    }
```
hoangnguyen92dn, line 34: Same here.

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
```diff
@@ -1,16 +1,34 @@
...
-)
+    @field: Json(name = "title") var title: String? = null,
+    @field: Json(name = "description") var description: String? = null,
+    @field: Json(name = "cover_image_url") var coverImageUrl: String? = null,
+    private var questions: HasMany<QuestionResponse>? = null,
+) : Resource() {
+
+    fun getQuestionResponses(): List<QuestionResponse>? {
+        val questions = questions?.get(document)
+        return if (questions?.firstOrNull() != null) questions else null
```
hoangnguyen92dn, line 19: As I can see, you have to validate this because of using same response for survey list and survey details, can you please create separate response and get rid of unnecessary validate here?

hoangnguyen92dn (commented):
@mikepham1011 @minhnimble Just one comment left 🚀

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyResponse.kt
minhnimble: oh understand your point, yeah I can definitely separate it, let me update 👍

minhnimble: I created two different responses but it triggers a crash issue, can't update ^^!

Screen Shot 2020-11-06 at 17 48 01

hoangnguyen92dn: Ahha. Gotta. Just double checked the survey details response, you need to update the current response class. We just need the list of `question` and `answer`, so that mean we could update to use `QuestionResponse` as the response for survey details endpoint, does it make sense to you? If so, can you please try with this solution?

minhnimble: okay let me try, if it works it is really nice lol

minhnimble: Hi @hoangnguyen92dn, I don't think it works :(( . When I tried to change the service to just provide the list of question responses only for loading survey details like this:
Screen Shot 2020-11-06 at 18 48 59

It still parse the object as type `survey` and the array of questions is now null:
Screen Shot 2020-11-06 at 18 48 48

hoangnguyen92dn: Hi @minhnimble Just did a small update and it's working as expected from my side, we need to consolidate the response classes and properties a little bit here:

1. `api/v1/surveys`

This endpoint returns `List<SurveyReponse>` with necessary data such as: title, description, cover image url

2. `/api/v1/surveys/{surveyId}`

This endpoint returns `List<SurveyDetailsResponse>` (SurveyDetailsResponse is the new class I created), we could handle to get the list of questions, answers from this endpoint.

Could you please try again from your side?

FYI: This is debugging result for temporary updated from my side

image

minhnimble: oh,thanks for the screenshots, it is clearer now, previously I thought you want to return the array of questions directly in the Service layer, but it should be in the Repository layer. Let me update

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyBasicResponse.kt
```diff
@@ -0,0 +1,26 @@
...
+data class SurveysMetaResponse(
+    @field: Json(name = "page") var page: Int? = null,
+    @field: Json(name = "pages") var pages: Int? = null
+)
+
+val List<SurveyBasicResponse>.meta: SurveysMetaResponse?
+    get() = when {
+        isEmpty() -> SurveysMetaResponse()
+        else -> first().document.meta.get<SurveysMetaResponse>(
+            Operators.moshi.adapter(SurveysMetaResponse::class.java)
```
hoangnguyen92dn, line 24: Hmm. Imo, define adapter for parsing data from response layer. Could we update to get rid of this? I guess we could define parent response class with `surveys` and `meta` 🤔

hoangnguyen92dn (commented):
In case we need to build for production with `minify` mode enabled, the app would be failed on launch because at least you are missing some configuration part on proguards for `banana` library. Please check and update if needed.

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyDetailsResponse.kt
```diff
@@ -0,0 +1,21 @@
...
+    @field: Json(name = "cover_image_url") var coverImageUrl: String? = null,
+    private var questions: HasMany<QuestionResponse>? = null,
+) : Resource() {
+
+    fun getQuestionResponses(): List<QuestionResponse>? {
+        return questions?.get(document)
+    }
+}
+
+
```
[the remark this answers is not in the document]
minhnimble: oh nice catch, let me update 👍

app/src/main/java/co/nimblehq/data/model/Survey.kt
```diff
@@ -1,30 +1,38 @@
...
-fun List<SurveyResponse>.toSurveys() = this.map { it.toSurvey() }
+fun List<SurveyBasicResponse>.toSurveys() = this.map { it.toSurvey() }
+
+fun SurveyDetailsResponse.toSurvey() = Survey(
+    description = description.orEmpty(),
+    id = id.orEmpty(),
+    imageUrl = coverImageUrl.orEmpty(),
+    title = title.orEmpty(),
+    questions = getQuestionResponses()?.toQuestions() ?: listOf()
+)
```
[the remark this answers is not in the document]
minhnimble: yep you are right, can remove I believe, let me check

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyBasicResponse.kt
minhnimble: This is the technique used in `The1` since we are using the JsonApi parser, not sure the technique you mention means going back to manual parsing? 🤔

hoangnguyen92dn: Hmm. My idea is get rid of the moshi adapter from Response class and define in the Factory same as `QuestionResponse` and `AnwserResponse`. Actually I am still ok with the current one solution, just wondering if we could revamp a little bit here.

hoangnguyen92dn (approved):
LGTM

data/src/main/java/co/nimblehq/data/api/response/survey/SurveyBasicResponse.kt
minhnimble: hmm, understood your point, I did experimenting a bit but it seems to not parse correctly. Not really sure how to do it right. Will definitely take a look again when we have more time 👌

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
```diff
@@ -2,13 +2,45 @@ package co.nimblehq.ui.screen.main.surveydetails
...
+        get() = _questionItemPagerUiModels
+
     val inputs: Inputs = this
+
+    fun loadSurveyDetails(surveyId: String) {
+        loadSurveyDetailsSingleUseCase
+            .execute(surveyId)
+            .doOnSubscribe { _showLoading.onNext(true) }
+            .doFinally { _showLoading.onNext(false) }
+            .subscribeBy (
```
manh-t, line 39: formatting 😳

manh-t (approved):
the PR looks better 💦  just a minor comment left 🚀

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
minhnimble: really nice catch, I missed it somehow 🤣 , will fix it 👍

PR #25, opened 2020-11-04
[TASK-10] [Logic] As a user, I can submit the survey after answers all the questions

minhnimble:
https://github.com/minhnimble/android-surveys-list/projects/1#card-46086096

## What happened

When completing answering all the questions, users will be able to submit the survey.

This is PR is for implementing the submit survey responses logic flow and connecting to submit survey response API.
 
## Insight

- [x] Implement submit survey logic using stored answers
- [x] Connect to submit survey API
 
## Proof Of Work
API submit survey response log:
Screen Shot 2020-11-04 at 11 03 21

Submit survey flow:
Screen Recording 2020-11-04 at 11 09 17

### DIFF
--- a/app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
+++ b/app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
@@ -2,16 +2,19 @@ package co.nimblehq.ui.screen.main.surveydetails
 
 import androidx.hilt.lifecycle.ViewModelInject
 import co.nimblehq.data.lib.common.DEFAULT_UNSELECTED_INDEX
+import co.nimblehq.data.model.QuestionResponsesEntity
 import co.nimblehq.extension.isValidIndex
 import co.nimblehq.ui.base.BaseViewModel
 import co.nimblehq.ui.screen.main.surveydetails.uimodel.QuestionItemPagerUiModel
 import co.nimblehq.ui.screen.main.surveydetails.uimodel.toQuestionItemPagerUiModels
 import co.nimblehq.usecase.survey.LoadSurveyDetailsSingleUseCase
+import co.nimblehq.usecase.survey.SubmitSurveyResponsesCompletableUseCase
 import io.reactivex.Observable
 import io.reactivex.rxkotlin.Observables
 import io.reactivex.rxkotlin.subscribeBy
 import io.reactivex.subjects.BehaviorSubject
 import io.reactivex.subjects.PublishSubject
+import java.util.concurrent.TimeUnit
 
 interface Inputs {
     fun currentQuestionIndex(index: Int)
@@ -20,12 +23,15 @@ interface Inputs {
 }
 
 class SurveyDetailsViewModel @ViewModelInject constructor(
-    private val loadSurveyDetailsSingleUseCase: LoadSurveyDetailsSingleUseCase
+    private val loadSurveyDetailsSingleUseCase: LoadSurveyDetailsSingleUseCase,
+    private val submitSurveyResponsesCompletableUseCase: SubmitSurveyResponsesCompletableUseCase
 ) : BaseViewModel(), Inputs {
 
     private val _showError = PublishSubject.create<Throwable>()
 
     private val _showLoading = BehaviorSubject.create<Boolean>()
+
+    private val _showSuccessOverlay = PublishSubject.create<Boolean>()
 
     private val _currentQuestionIndex = BehaviorSubject.create<Int>()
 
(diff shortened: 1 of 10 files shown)


app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
```diff
@@ -89,5 +98,19 @@ class SurveyDetailsViewModel @ViewModelInject constructor(
...
     }
+
+    fun submitSurveyResponses(surveyId: String, responses: List<QuestionResponsesEntity>) {
+        submitSurveyResponsesCompletableUseCase
+            .execute(SubmitSurveyResponsesCompletableUseCase.Input(surveyId, responses))
+            .doOnSubscribe { _showLoading.onNext(true) }
+            .doFinally { _showLoading.onNext(false) }
+            .doOnComplete { _showSuccessOverlay.onNext(true) }
+            .doOnError { _showError.onNext(it) }
+            .delay(3, TimeUnit.SECONDS)
```
manh-t, line 109: any reasons why we need to delay 3 sec? 🤓

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
```diff
@@ -89,5 +98,19 @@ class SurveyDetailsViewModel @ViewModelInject constructor(
             )
             .bindForDisposable()
     }
+
+    fun submitSurveyResponses(surveyId: String, responses: List<QuestionResponsesEntity>) {
+        submitSurveyResponsesCompletableUseCase
+            .execute(SubmitSurveyResponsesCompletableUseCase.Input(surveyId, responses))
+            .doOnSubscribe { _showLoading.onNext(true) }
+            .doFinally { _showLoading.onNext(false) }
+            .doOnComplete { _showSuccessOverlay.onNext(true) }
```
manh-t, line 107: hmm, you are using both `doOnComplete` and `subscribe(onComplete=..)`. Could you please explain why we need `doOnComplete` here? 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
```diff
@@ -89,5 +98,19 @@ class SurveyDetailsViewModel @ViewModelInject constructor(
...
+            .doOnError { _showError.onNext(it) }
```
manh-t, line 108: Optional: I think we can put `_showError` to `onError` in subscribe to take the advantage of `subscribeBy` method. 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
minhnimble: This is for testing purpose that it will represent the upcoming lottie animation (not implemented yet) and now we just see a toast instead for this period 🙏. Please ignore here since the next branch I will replace it with the actual lottie animation and remove the delay 😊

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
minhnimble: This is for testing purpose to represent the lottie animation time, I will update this place to correct behavior when we have the actually lottie animation in the up-coming branch, please ignore this flow for now 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
minhnimble: Oh if there is an error after running the use case, then if we put `_showError` to `onError` in subscribe, wouldn't it delay 3 seconds then fire the error? I want the flow to fire the error as soon as the use case fails, that why I did it like this 😊. What do you think? 🤔

manh-t: @minhnimble okay, I think this flow would be changed one more time with the full implementation. Let's see 👍

minhnimble: yeah @manh-t I will update it to the correct flow in the next PR, let's go 👍

minhnimble:
Hi @manh-t, I just rebased and updated this branch to latest master, you can take a look again now, thank youu 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsViewModel.kt
minhnimble: Added a TODO for explaining about the hard-coded delay period 🙏

manh-t (approved):
Sorry for late 🙇  LGTM 🚀

sleepylee (approved):

PR #27, opened 2020-11-04
[TASK-11] [UI] As a user, I can see an animation on completing the survey

minhnimble:
https://github.com/minhnimble/android-surveys-list/projects/1#card-46087643

## What happened

When completing a survey, users will be able to see a Lottie-based animation (or alternative) on completing the survey, then going back to surveys list screen.

This PR is for creating complete survey Lottie-based animation and go back to surveys list flow after showing the animation 🎨.

**Design**
[Complete survey animation](https://www.figma.com/proto/GjRPOjDyZ6f4EDL3wKarRK/Challenge---Mobile-App?node-id=38%3A2462&scaling=scale-down)
 
## Insight

- [x] Implement complete survey animation
- [x] Implement go back to surveys list flow after animation ended
 
## Proof Of Work
Screen Recording 2020-11-04 at 13 58 29


app/src/main/java/co/nimblehq/ui/common/dialog/ConfirmExitSurveyDialog.kt
```diff
@@ -24,7 +25,7 @@ class ConfirmExitSurveyDialog(private val confirmCallback: () -> Unit): DialogFr
         }
         view.tvConfirmExitSurveyYes.setOnClickListener {
             dismiss()
-            confirmCallback.invoke()
+            confirmCallback()
```
manh-t, line 28: oh, I'm curious if `confirmCallback()` == `confirmCallback.invoke()`. Is it correct? I see you are using old way in `SubmitSurveyResponsesSuccessDialog`

manh-t (commented):
This implementation makes more sense to me 👍

app/src/main/java/co/nimblehq/ui/common/dialog/ConfirmExitSurveyDialog.kt
minhnimble: yep it is correct that `confirmCallback() == confirmCallback.invoke()` 😊. I still use the old way in `SubmitSurveyResponsesSuccessDialog` because the `callback` is nullable, and to trigger it safely we need to use `invoke` here. I can't just do `callback?()` it will prompt a syntax error so I have to keep it 🙏

manh-t: thank you for enlightening me 💡

manh-t (approved):

app/src/main/java/co/nimblehq/ui/common/dialog/SubmitSurveyResponsesSuccessDialog.kt
```diff
@@ -0,0 +1,31 @@
...
+import kotlin.concurrent.schedule
+
+class SubmitSurveyResponsesSuccessDialog(private val callback: EmptyCallback? = null) : DialogFragment() {
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        val dialog = Dialog(context, R.style.CenterFadeInOutDialogStyle)
+        dialog.window?.setBackgroundDrawable(ColorDrawable(android.graphics.Color.TRANSPARENT))
+        val view = LayoutInflater.from(context).inflate(R.layout.dialog_submit_survey_responses_success, null)
+        dialog.setContentView(view)
+        return dialog
```
sleepylee, line 21: to implement a `DialogFragment`, I'd suggest you go through the same lifecycle as a normal fragment.
but if you would like to use this onCreateDialog callback I'd suggest rearranging a bit:

```suggestion
val view = LayoutInflater.from(context).inflate(R.layout.dialog_submit_survey_responses_success, null)
return Dialog(context, R.style.CenterFadeInOutDialogStyle).apply {
    window?.setBackgroundDrawable(ColorDrawable(android.graphics.Color.TRANSPARENT))
    setContentView(view)
}
```

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsFragment.kt
```diff
@@ -121,10 +122,9 @@ class SurveyDetailsFragment : BaseFragment(), BaseFragmentCallbacks {
 
     private fun bindShowSuccessOverlay(shouldShow: Boolean) {
         if (shouldShow) {
-            // TODO: Handle show success lottie overlay
-            displayError(AppError(null, "Submit survey responses successfully."))
-        } else {
-            navigator.navigateBack()
+            SubmitSurveyResponsesSuccessDialog {
+                navigator.navigateBack()
+            }.show(childFragmentManager, "success_animation")
```
sleepylee, line 127: I think you won't need the tag, but if you do, you may want a const

sleepylee (approved):
minor things and we can go

app/src/main/java/co/nimblehq/ui/common/dialog/SubmitSurveyResponsesSuccessDialog.kt
minhnimble: thank you for your suggestion, will update 🙏

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsFragment.kt
minhnimble: hmm, as of now the tag in place is just for future identification, it is not being used at the moment, I think I can remove it for now and will add it as const when need it in the future

PR #30, opened 2020-11-05
[TASK-13] [UI + Logic] As a user, I can see the user's info after logging in successfully

minhnimble:
https://github.com/minhnimble/android-surveys-list/projects/1#card-47407830

## What happened

When logged in successfully, user will be navigated to home Surveys List Screen. In here they can also see their avatar and name as well via right Menu.

**Design**
[Surveys list UI with avatar](https://www.figma.com/proto/GjRPOjDyZ6f4EDL3wKarRK/Challenge---Mobile-App?node-id=30%3A46&scaling=scale-down)
[Right Menu UI with avatar and name](https://www.figma.com/proto/GjRPOjDyZ6f4EDL3wKarRK/Challenge---Mobile-App?node-id=30%3A46&scaling=min-zoom)
 
## Insight

- [x] Connect to user info API
- [x] Implement get user info logic in the background
- [x] Update the avatar UI with actual user image instead of using a default one as of now in the Surveys List Screen and Right Menu Drawer.
- [x] Update the user's name in the Right Menu Drawer.

## Proof Of Work

API get user info success response log:
Screen Shot 2020-11-05 at 16 59 16

Display user data in right menu:

### DIFF
--- a/app/src/main/java/co/nimblehq/data/model/User.kt
+++ b/app/src/main/java/co/nimblehq/data/model/User.kt
@@ -1,0 +1,15 @@
+package co.nimblehq.data.model
+
+import co.nimblehq.data.api.response.user.UserResponse
+
+data class User(
+    var id: String = "",
+    var email: String = "",
+    var avatarUrl: String = ""
+)
+
+fun UserResponse.toUser() = User(
+    id = id.orEmpty(),
+    email = email.orEmpty(),
+    avatarUrl = avatarUrl.orEmpty()
+)
--- a/app/src/main/java/co/nimblehq/data/model/User.kt
+++ b/app/src/main/java/co/nimblehq/data/model/User.kt
@@ -3,9 +3,9 @@ package co.nimblehq.data.model
 import co.nimblehq.data.api.response.user.UserResponse
 
 data class User(
-    var id: String = "",
-    var email: String = "",
-    var avatarUrl: String = ""
+    var id: String,
+    var email: String,
+    var avatarUrl: String
 )
 
 fun UserResponse.toUser() = User(
--- a/app/src/main/java/co/nimblehq/di/modules/RetrofitModule.kt
+++ b/app/src/main/java/co/nimblehq/di/modules/RetrofitModule.kt
@@ -5,6 +5,7 @@ import co.nimblehq.data.api.providers.ConverterFactoryProvider
 import co.nimblehq.data.api.providers.RetrofitProvider
 import co.nimblehq.data.api.service.auth.AuthService
 import co.nimblehq.data.api.service.survey.SurveyService
+import co.nimblehq.data.api.service.user.UserService
 import co.nimblehq.di.qualifier.*
 import com.squareup.moshi.Moshi
 import dagger.Module
@@ -61,4 +62,8 @@ class RetrofitModule {
     @Provides
     @Singleton
     fun provideSurveyService(@AppRetrofit retrofit: Retrofit): SurveyService = ApiServiceProvider.getSurveyService(retrofit)
+
+    @Provides
+    @Singleton
+    fun provideUserService(@AppRetrofit retrofit: Retrofit): UserService = ApiServiceProvider.getUserService(retrofit)
 }
--- a/app/src/main/java/co/nimblehq/extension/ImageViewExtension.kt
+++ b/app/src/main/java/co/nimblehq/extension/ImageViewExtension.kt
@@ -1,5 +1,6 @@
 package co.nimblehq.extension
 
+import android.content.Context
 import android.graphics.drawable.ColorDrawable
 import android.graphics.drawable.Drawable
 import android.widget.ImageView
@@ -16,7 +17,7 @@ import com.bumptech.glide.load.resource.drawable.DrawableTransitionOptions
 fun ImageView.loadImage(url: String, placeHolderDrawable: Drawable? = null) {
     GlideApp.with(context)
         .load(url)
-        .placeholder(placeHolderDrawable ?: ColorDrawable(ContextCompat.getColor(context, R.color.black_20a)))
+        .placeholder(placeHolderDrawable ?: ColorDrawableCreator.default(context))
         .diskCacheStrategy(DiskCacheStrategy.ALL)
         .centerCrop()
         .into(this)
@@ -25,10 +26,14 @@ fun ImageView.loadImage(url: String, placeHolderDrawable: Drawable? = null) {
 fun ImageView.loadImageWithFadeAnimation(url: String, placeHolderDrawable: Drawable? = null) {
     GlideApp.with(context)
         .load(url)
-        .placeholder(placeHolderDrawable ?: ColorDrawable(ContextCompat.getColor(context, R.color.black_20a)))
+        .placeholder(placeHolderDrawable ?: ColorDrawableCreator.default(context))
         .transition(DrawableTransitionOptions.withCrossFade(1000))
         .diskCacheStrategy(DiskCacheStrategy.ALL)
         .centerCrop()
         .into(this)
+}
 
+object ColorDrawableCreator {
+    fun default(context: Context): ColorDrawable =
+        ColorDrawable(ContextCompat.getColor(context, R.color.black_20a))
 }
--- a/app/src/main/java/co/nimblehq/extension/ImageViewExtension.kt
+++ b/app/src/main/java/co/nimblehq/extension/ImageViewExtension.kt
@@ -1,6 +1,7 @@
 package co.nimblehq.extension
 
 import android.graphics.drawable.ColorDrawable
+import android.graphics.drawable.Drawable
 import android.widget.ImageView
 import androidx.core.content.ContextCompat
 import co.nimblehq.R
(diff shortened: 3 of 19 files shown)


app/src/main/java/co/nimblehq/extension/ImageViewExtension.kt
```diff
@@ -12,19 +13,19 @@ import com.bumptech.glide.load.resource.drawable.DrawableTransitionOptions
...
         .centerCrop()
         .into(this)
 }
 
-fun ImageView.loadImageWithFadeAnimation(url: String) {
+fun ImageView.loadImageWithFadeAnimation(url: String, placeHolderDrawable: Drawable? = null) {
     GlideApp.with(context)
         .load(url)
-        .placeholder(ColorDrawable(ContextCompat.getColor(context, R.color.black_20a)))
+        .placeholder(placeHolderDrawable ?: ColorDrawable(ContextCompat.getColor(context, R.color.black_20a)))
```
sleepylee, line 28: repetitive `ColorDrawable(ContextCompat.getColor(context, R.color.black_20a))` in 2 functions maybe you want to create a default `val`

app/src/main/java/co/nimblehq/ui/screen/main/MainActivity.kt
```diff
@@ -22,12 +25,26 @@ class MainActivity : BaseActivity(), LoaderAnimatable {
...
+
         setContentView(R.layout.activity_main)
         toggleShimmerLoader(true)
+        bindViewModel()
+    }
+
+    override fun onPostCreate(savedInstanceState: Bundle?) {
+        super.onPostCreate(savedInstanceState)
+
+        viewModel.loadCurrentUserInfo()
```
sleepylee, line 37: better move this to viewModel init(): onPostCreate is not very common, but I think basically if the Activity is recreated by rotating the screen, then it will be invoked as well, and this function is called - which is not necessary. Bring it to ViewModel so you just make sure the call is made once and the data is persisted without bothering the View's lifecycle

sleepylee: also, `init()` of ViewModel will be invoked as soon as the ViewModel is created as well, so when View is ready, sometimes we can have the data loaded faster before you start binding the UI

app/src/main/java/co/nimblehq/ui/screen/main/MainViewModel.kt
```diff
@@ -1,10 +1,32 @@
...
+
+    val error: Observable<Throwable>
+        get() = _error
+
+    fun loadCurrentUserInfo() {
+        loadCurrentUserInfoSingleUseCase
+            .execute(Unit)
+            .subscribeBy(
+                onError = { _error.onNext(it) },
+                onSuccess = { RxBus.publish(it) }
```
sleepylee, line 29: don't use this way, I had experienced with it before, it's very convenient at first, but when you need to debug or modify the code it's very annoyed to find the source of emission. Instead, you can create a BehaviorSubject inside a new UserViewModel() - which is used to maintain the User's state. You can still have this MainViewModel to manage the state of the View as well. In general, you can have 2 ViewModel in 1 View, that's ok. And that UserViewModel can be used in other places - for example: calling from Fragment to get the current UserProfile or UserState. By subscribing and dispose when the View is destroyed, you are committing to better resource usage and release when not needed, avoid leaking memory, overall improving the app performance

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
```diff
@@ -106,6 +114,13 @@ class SurveysFragment: BaseFragment(), BaseFragmentCallbacks, NavigationView.OnN
         return false
     }
 
+    private fun bindCurrentUserInfo(user: User) {
+        tvMenuDrawerUserName.text = user.email
+        val defaultAvatar = requireContext().resources.getDrawable(R.drawable.ic_general_default_user_avatar, requireContext().theme)
```
sleepylee, line 119: I think it's safer and a new way to introduce you with `ContextCompat. getDrawable`, sometimes there are projects that support lower API like < 21, some of the methods to get resources required extra params, or deprecated, or longer to write. In this case, you can write this as short and safe like this:
`ContextCompat.getDrawable(requireContext(), R.drawable.your_image)`

data/src/main/java/co/nimblehq/data/api/service/user/UserService.kt
```diff
@@ -0,0 +1,16 @@
+package co.nimblehq.data.api.service.user
+
+import co.nimblehq.data.api.response.user.UserResponse
+import io.reactivex.Flowable
+import retrofit2.Response
+import retrofit2.http.GET
+
+
```
sleepylee, line 8: space

app/src/main/java/co/nimblehq/ui/screen/main/surveys/SurveysFragment.kt
minhnimble: thank you for your suggestion, will update 🙏

app/src/main/java/co/nimblehq/extension/ImageViewExtension.kt
minhnimble: thanks for spotting, let me update

app/src/main/java/co/nimblehq/ui/screen/main/MainActivity.kt
minhnimble: your suggestion make senses, let me handle it in `init()`

app/src/main/java/co/nimblehq/ui/screen/main/MainViewModel.kt
minhnimble: noted, understood your point, let me update it 👍

minhnimble:
@sleepylee, all comments are fixed, you can check again now 🙏

app/src/main/java/co/nimblehq/data/model/User.kt
```diff
@@ -0,0 +1,15 @@
+package co.nimblehq.data.model
+
+import co.nimblehq.data.api.response.user.UserResponse
+
+data class User(
+    var id: String = "",
+    var email: String = "",
+    var avatarUrl: String = ""
```
manh-t, line 8: do we need default value here for any specific case? 🙏

manh-t (approved):
just a minor question, rest LGTM

app/src/main/java/co/nimblehq/data/model/User.kt
minhnimble: oh in the case we want to create a user object with simply this `User()` for testing purpose. You can see it more clearly in the test branch. At this moment in this branch you won't be able to see it yet 🙏

manh-t: @minhnimble yeah I see, but I think we should write test without affecting the logic 😄  I think we can create the default value in the test instead. How do you think? 🙏

minhnimble: I think that is okay too, let me update 👍

app/src/main/java/co/nimblehq/di/modules/RetrofitModule.kt
```diff
@@ -61,4 +62,8 @@ class RetrofitModule {
     @Provides
     @Singleton
     fun provideSurveyService(@AppRetrofit retrofit: Retrofit): SurveyService = ApiServiceProvider.getSurveyService(retrofit)
+
+    @Provides
+    @Singleton
```
sleepylee, line 67: be very careful with @Singleton 😉  later if you have time you can research this and should update it to a smaller scope if possible, depending on each component usage

app/src/main/res/layout/navigation_drawer_header_surveys.xml
```diff
@@ -1,48 +1,47 @@
...
-      app:layout_constraintEnd_toEndOf="parent"
-      app:layout_constraintStart_toStartOf="parent">
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:id="@+id/clMenuDrawerSeparatorLine"
+        android:layout_width="match_parent"
+        android:layout_height="0.5dp"
+        android:background="@color/white_18a"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent">
```
sleepylee, line 43: - if you just want to draw, then use View, it's smaller in term of creation
- if you match parent, then you don't need constraint start + end
- close the tag if you don't have any child item

sleepylee (approved):
some minor points, but should be good to go after fixing them

app/src/main/java/co/nimblehq/di/modules/RetrofitModule.kt
minhnimble: noted 🙏

app/src/main/res/layout/navigation_drawer_header_surveys.xml
minhnimble: will update to use view for this separator and update its constraints, this is a nice catch 👍

sleepylee: yeah but you can see that there is no error on the layout editor, actually, I think it's just recommended, but yeah 0dp + start/end constraint is better, because then you can do margin if you want (match_parent can't)

PR #31, opened 2020-11-05
[TASK-13] [Test] As a user, I can see the user's info after logging in successfully

minhnimble:
https://github.com/minhnimble/android-surveys-list/projects/1#card-47407830

## What happened

Need to write unit tests for classes used in load user info feature.
 
## Insight

**Acceptance Criteria**
- [x] Write unit tests
 
## Proof Of Work
No visible changes, just under the hood tests

### DIFF
--- a/app/src/main/java/co/nimblehq/ui/screen/common/UserViewModel.kt
+++ b/app/src/main/java/co/nimblehq/ui/screen/common/UserViewModel.kt
@@ -32,7 +32,7 @@ class UserViewModel @ViewModelInject constructor(
         loadCurrentUserInfo()
     }
 
-    private fun loadCurrentUserInfo() {
+    fun loadCurrentUserInfo() {
         loadCurrentUserInfoSingleUseCase
             .execute(Unit)
             .subscribeBy(
--- a/app/src/main/java/co/nimblehq/ui/screen/common/UserViewModel.kt
+++ b/app/src/main/java/co/nimblehq/ui/screen/common/UserViewModel.kt
@@ -6,6 +6,7 @@ import co.nimblehq.ui.base.BaseViewModel
 import co.nimblehq.usecase.user.LoadCurrentUserInfoSingleUseCase
 import io.reactivex.Observable
 import io.reactivex.rxkotlin.subscribeBy
+import io.reactivex.subjects.BehaviorSubject
 import io.reactivex.subjects.PublishSubject
 
 interface Output {
@@ -24,7 +25,7 @@ class UserViewModel @ViewModelInject constructor(
     override val error: Observable<Throwable>
         get() = _error
 
-    private val _user = PublishSubject.create<User>()
+    private val _user = BehaviorSubject.create<User>()
     override val user: Observable<User>
         get() = _user
 
@@ -32,7 +33,7 @@ class UserViewModel @ViewModelInject constructor(
         loadCurrentUserInfo()
     }
 
-    fun loadCurrentUserInfo() {
+    private fun loadCurrentUserInfo() {
         loadCurrentUserInfoSingleUseCase
             .execute(Unit)
             .subscribeBy(
--- a/app/src/test/java/co/nimblehq/ui/screen/common/UserViewModelTest.kt
+++ b/app/src/test/java/co/nimblehq/ui/screen/common/UserViewModelTest.kt
@@ -10,6 +10,7 @@ import com.nhaarman.mockitokotlin2.whenever
 import io.reactivex.Single
 import org.junit.Before
 import org.junit.Test
+import org.robolectric.util.ReflectionHelpers
 
 class UserViewModelTest {
 
@@ -35,7 +36,7 @@ class UserViewModelTest {
             .output
             .error
             .test()
-        userViewModel.loadCurrentUserInfo()
+        ReflectionHelpers.callInstanceMethod<UserViewModel>(userViewModel, "loadCurrentUserInfo")
 
         // Assert
         errorObserver
@@ -58,7 +59,6 @@ class UserViewModelTest {
             .output
             .user
             .test()
-        userViewModel.loadCurrentUserInfo()
 
         // Assert
         userObserver
--- a/app/src/test/java/co/nimblehq/ui/screen/common/UserViewModelTest.kt
+++ b/app/src/test/java/co/nimblehq/ui/screen/common/UserViewModelTest.kt
@@ -1,0 +1,71 @@
+package co.nimblehq.ui.screen.common
+
+import co.nimblehq.data.error.UserError
+import co.nimblehq.data.model.User
+import co.nimblehq.usecase.user.LoadCurrentUserInfoSingleUseCase
+import com.nhaarman.mockitokotlin2.any
+import com.nhaarman.mockitokotlin2.doReturn
+import com.nhaarman.mockitokotlin2.mock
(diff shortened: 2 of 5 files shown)


app/src/test/java/co/nimblehq/ui/screen/common/UserViewModelTest.kt
```diff
@@ -0,0 +1,71 @@
...
+            mockLoadCurrentUserInfoSingleUseCase.execute(any())
+        ) doReturn Single.error(UserError.LoadCurrentUserInfoError(null))
+
+        // Act
+        userViewModel = UserViewModel(mockLoadCurrentUserInfoSingleUseCase)
+        val errorObserver = userViewModel
+            .output
+            .error
+            .test()
+        userViewModel.loadCurrentUserInfo()
```
manh-t, line 38: oh, I think when we initialize VM, the code in `init` block should be auto called, right? If so, we won't need to make public for `loadCurrentUserInfo()`, and make sure that we mock `usecase`'s behavior before the function call. Could you please recheck this? 🙏

minhnimble: > oh, I think when we initialize VM, the code in init block should be auto called, right? If so, we won't need to make public for loadCurrentUserInfo()

very nice concern, this is something I actually have trouble to test it and I have to public `loadCurrentUserInfo()` to trigger the error, otherwise the test will fail even though it is called in the `init` block. Not sure if there is a better suggestion.

> make sure that we mock usecase's behavior before the function call. Could you please recheck this? 🙏

oh right nice catch, I was trying to trigger the usecase with `init` but it doesn't work and I forget to revert. Let me update 👍

minhnimble:
@manh-t, please help me take a look again, thank youu 🙏

app/src/test/java/co/nimblehq/ui/screen/common/UserViewModelTest.kt
manh-t: hmm yeah, it might be the race condition between creating VM and observing the observable, so we cannot capture the value 🤔

manh-t: > oh I think usecase's behavior is mocked before the function call. Not sure what do you mean 🤔

ah, nevermind, my bad. The flow is correct 😓

manh-t: okay, we can temporarily keep it as it is. I will take note and find the solution later. 😁

manh-t (approved):

app/src/main/java/co/nimblehq/ui/screen/common/UserViewModel.kt
```diff
@@ -32,7 +32,7 @@ class UserViewModel @ViewModelInject constructor(
         loadCurrentUserInfo()
     }
 
-    private fun loadCurrentUserInfo() {
+    fun loadCurrentUserInfo() {
```
sleepylee, line 35: why? 🤔  you just need to instantiate the ViewModel, this `loadCurrentUserInfo()` will be covered in `init { }` so you won't have to call it manually, just test the emitter it sends the value to, .a.k.a the BehaviorSubject

minhnimble: I tried the tests with `ViewModel`'s init already, and the emitter doesn't receive any value, that is the weird part and I have to use this way to perform the actual trigger 🙏

This is how the returned results look like when testing with just `ViewModel`'s init and not calling the `loadCurrentUserInfo()`:
Screen Shot 2020-12-02 at 10 38 07

hoangnguyen92dn: @minhnimble As I see, you are using `PublishSubject` for `output.user`, by using `PublishSubject` you only receive the emitted one once subscribed (Refer: http://reactivex.io/RxJava/javadoc/rx/subjects/PublishSubject.html). So that's why Trung suggest to use `BehaviorSubject` (Refer: http://reactivex.io/RxJava/javadoc/io/reactivex/subjects/BehaviorSubject.html)

minhnimble: @hoangnguyen92dn, thanks for spotting out that right now we are using `PublishSubject`, I am totally not aware, let me update 👍

minhnimble:
Hi @sleepylee and @hoangnguyen92dn, I updated the branch to reflect the correct tests when `loadCurrentUserInfo` is called inside `init` function. You guys can check again now, thanks 🙏

app/src/main/java/co/nimblehq/ui/screen/common/UserViewModel.kt
```diff
@@ -20,11 +20,11 @@ class UserViewModel @ViewModelInject constructor(
 
     val output = this
 
-    private val _error = PublishSubject.create<Throwable>()
+    private val _error = BehaviorSubject.create<Throwable>()
```
hoangnguyen92dn, line 23: Why do we need to update `_error` to `BehaviorSubject`?

minhnimble: same reason with the `output.user` Observable 😅

hoangnguyen92dn: We should keep the `_error` as `PublishSubject`, otherwise, let imagine when an error emitted (and shown message on toast) then we subscribe the `error` from Fragment again (for some reason, navigate to next screen then back, ..etc..), the error will be shown again because the last emitted still stored in `BehaviorSubject` and will be notified on subscribe immediately.

minhnimble: @hoangnguyen92dn noted, thanks for your suggestion on the Reflection capability to test private methods. I have fixed the tests and they are working now. You can check again 🙏

hoangnguyen92dn: You did a wrong revert 😅  `_error` should be `PublishSubject` and `_user` should be `BehaviorSubject`

app/src/test/java/co/nimblehq/ui/screen/common/UserViewModelTest.kt
```diff
@@ -0,0 +1,72 @@
...
+
+        // Act
+        userViewModel = UserViewModel(mockLoadCurrentUserInfoSingleUseCase)
+        val userObserver = userViewModel
+            .output
+            .user
+            .test()
+        val loadCurrentUserInfoMethod = userViewModel.javaClass.getDeclaredMethod("loadCurrentUserInfo")
+        loadCurrentUserInfoMethod.isAccessible = true
+        loadCurrentUserInfoMethod.invoke(userViewModel)
```
hoangnguyen92dn, line 62: It would be more simple by using `ReflectionHelpers.callInstanceMethod()` from Robolectric

minhnimble: oh this is so cool to know, let me update 👍

hoangnguyen92dn (approved):
LGTM

PR #32, opened 2020-11-23
[Chores] Improve view models architecture by adding Output interfaces for observables

minhnimble:
## What happened

Just a small refactor work regarding the ViewModels incoming and outgoing communications, update Production's clientId & secrets and fix some warnings. To add more context:

- Implemented Output interfaces for ViewModels, help group output Observables for better code structuring (similar with DeeMoney Android project)
 - Add `Locale.getDefault()` to `toUpperCase` function for strings to remove warnings.
- Update app with valid `ClientId` and `Secret` for OAuth2 authentication with Production server.
 
## Proof Of Work

The application should be built and functioning normally. Just code refactor and no visible changes.

### DIFF
--- a/app/src/main/java/co/nimblehq/data/model/Question.kt
+++ b/app/src/main/java/co/nimblehq/data/model/Question.kt
@@ -1,6 +1,7 @@
 package co.nimblehq.data.model
 
 import co.nimblehq.data.api.response.survey.QuestionResponse
+import java.util.*
 
 enum class QuestionDisplayType {
     INTRO,
@@ -21,7 +22,7 @@ enum class QuestionDisplayType {
         fun from(value: String?): QuestionDisplayType {
             value ?: return DEFAULT
             return try {
-                valueOf(value.toUpperCase())
+                valueOf(value.toUpperCase(Locale.ROOT))
             } catch (ex: Exception) {
                 DEFAULT
             }
@@ -38,7 +39,7 @@ enum class QuestionPickValue {
         fun from(value: String?): QuestionPickValue {
             value ?: return NONE
             return try {
-                valueOf(value.toUpperCase())
+                valueOf(value.toUpperCase(Locale.ROOT))
             } catch (ex: Exception) {
                 NONE
             }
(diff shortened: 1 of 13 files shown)


app/src/main/java/co/nimblehq/data/model/Question.kt
```diff
@@ -21,7 +22,7 @@ enum class QuestionDisplayType {
         fun from(value: String?): QuestionDisplayType {
             value ?: return DEFAULT
             return try {
-                valueOf(value.toUpperCase())
+                valueOf(value.toUpperCase(Locale.ROOT))
```
sleepylee, line 25: I think it's better to be `defaultLocale()`, or you should pass the `Locale` in the params, the problem is that sometimes, when supporting switching locale in an application, the application locale won't change, but the activity's locale will change, in that case, to get the correct value, you must pass the Locale from the Activity context.

Since this is just an UpperCase func, I think we can just use defaultLocale()

app/src/main/java/co/nimblehq/ui/screen/main/surveydetails/SurveyDetailsFragment.kt
```diff
@@ -61,7 +62,7 @@ class SurveyDetailsFragment : BaseFragment(), BaseFragmentCallbacks {
         }.bindForDisposable()
 
         btSurveyQuestionsItemNext.subscribeOnClick {
-            viewModel.inputs.triggerNextQuestion()
+            viewModel.input.triggerNextQuestion()
```
sleepylee, line 65: I think plural `inputs` is ok too, because we have many of them, but this is fine as well because we follow the class Name

sleepylee (approved):
minor comments, then we go

app/src/main/java/co/nimblehq/data/model/Question.kt
minhnimble: thank you for your insights, at this place I think we just need to use `defaultLocale()` is enough 👍

manh-t (approved):

PR #33, opened 2020-11-23
[TASK-14] [Logic] As a user, I can see the app automatically refresh the user token in an active session if needed

minhnimble:
https://github.com/minhnimble/internal-certification-surveys-android/projects/1#card-49875069

## What happened

When user is using the app and is on an active logged in session, then if the token expires, the app need to be able to automatically refresh the token in the background, or kick the user out of this session back to Login Screen if the app can't refresh that token.
 
## Insight

- [x] Implement refresh token logic in the background while the user is on an active session and the token reach its expired time
- [x] Implement kick user out of this session to Login Screen logic if the app can't refresh the token on an active session

## Proof Of Work

This is hard to test since we need to rely on the back-end responses 😅. The real way for testing it is after the first login session, wait until it almost reaches `7200` seconds mark then start the app with that session. From there after the time passes the `7200` seconds mark, trigger any API call to trigger the refresh token flow.

One way I am testing it right now is after a fresh login, instead of sending a correct token for getting surveys list when we are at Home Surveys Screen, we will send a wrong token to backend to trigger the refresh token flow. This would require some hard-coding. Here is the log that represents that flow:

[log_sequences_refreshing_tokens.txt](https://github.com/minhnimble/internal-certification-surveys-android/files/5588053/log_sequences_refreshing_tokens.txt)


data/src/main/java/co/nimblehq/data/api/interceptor/AppRequestInterceptor.kt
```diff
@@ -1,22 +1,69 @@
...
-    private val securedStorage: SecureStorage
+    private val securedStorage: SecureStorage,
+    private val authService: AuthService
 ) : Interceptor {
 
     @Throws(IOException::class)
     override fun intercept(chain: Interceptor.Chain): Response {
-        val request = chain.request().newBuilder()
+        val response = executeRequest(chain)
+        if (response.code == 401) { // Current token is expired, trigger refresh token API call to get a new token
```
sleepylee, line 25: checkout `Authenticator`, what it does is that it will invoke a callback whenever facing 401 automatically for you, so you don't have to handle this part by yourself, then you can refresh the token, while still holding the same previous request before, retry to execute it with the newly obtained token. You can check in The 1 for reference.
Check this post out: 
https://medium.com/knowing-android/headers-interceptors-and-authenticators-with-retrofit-1a00fed0d5eb

minhnimble: will take a look and improve 🙏 previously I followed this post to implement it 😅
https://gist.github.com/alex-shpak/da1e65f52dc916716930

minhnimble:
@sleepylee, all comments are fixed, you can check again now 🙏

app/src/main/java/co/nimblehq/ui/screen/main/MainActivity.kt
```diff
@@ -25,10 +27,22 @@ class MainActivity : BaseActivity(), LoaderAnimatable {
...
 
     override fun toggleShimmerLoader(shouldShow: Boolean) {
         clMainShimmerContainer.visibility = if (shouldShow) View.VISIBLE else View.GONE
         if (shouldShow) sflMainContainer.startShimmer() else sflMainContainer.stopShimmer()
     }
+
+    private fun bindViewModel() {
+        RxBus.listen(TokenExpiredEvent::class.java)
+            .subscribe { showOnboardingActivity() }
+            .bindForDisposable()
```
manh-t, line 41: as Trung mentioned before, I think we shouldn't use this approach to listen when refreshing token failed. Instead, you can create an interface with listener to cover this case. You can reference to The 1 to see how it works 😁

minhnimble: yeah sure, let's me investigate 🙏

minhnimble:
Hi @manh-t, I updated the branch with the new Listener like The1 instead of RxBus, you can take a look now 🙏

data/src/main/java/co/nimblehq/data/api/interceptor/AppRequestInterceptor.kt
```diff
@@ -1,22 +1,77 @@
...
+     * Authenticator for when the authToken need to be refresh and updated
+     * every time we get a 401 error code
+     */
+    @Throws(IOException::class)
+    override fun authenticate(route: Route?, response: Response): Request? {
+        val refreshTokenRequest = RequestHelper.refreshToken(
+            securedStorage.userRefreshToken ?: ""
+        )
+        return try {
+            val refreshTokenResponse = authService.refreshToken(refreshTokenRequest)
```
sleepylee, line 37: should have a retry time track, otherwise if API always returns 401 for an invalid auth there will be a recursive loop

data/src/main/java/co/nimblehq/data/api/interceptor/AppRequestInterceptor.kt
```diff
@@ -1,22 +1,77 @@
...
+import okhttp3.*
 import java.io.IOException
 import javax.inject.Inject
 
 class AppRequestInterceptor @Inject constructor(
-    private val securedStorage: SecureStorage
-) : Interceptor {
+    private val securedStorage: SecureStorage,
+    private val authService: AuthService
+) : Interceptor, Authenticator {
```
sleepylee, line 17: why not separating them? 🤔  they are doing 2 different things, always remember `composition over inherritance`

data/src/main/java/co/nimblehq/data/api/interceptor/AppRequestInterceptor.kt
minhnimble: noted, let me update 👍

data/src/main/java/co/nimblehq/data/api/interceptor/AppRequestInterceptor.kt
minhnimble: in the example here: https://medium.com/knowing-android/headers-interceptors-and-authenticators-with-retrofit-1a00fed0d5eb, it seems to be okay to combine them. However, I think your idea definitely make more sense to me. Sure will update

minhnimble:
@sleepylee, all comments are fixed, you can check again now 🙏

data/src/main/java/co/nimblehq/data/api/interceptor/AppRequestAuthenticator.kt
```diff
@@ -0,0 +1,83 @@
...
+) : Authenticator {
+
+    private var refreshTokenRetryTime = 1
+
+    /**
+     * Authenticator for when the authToken need to be refresh and updated
+     * every time we get a 401 error code
+     */
+    @Throws(IOException::class)
+    override fun authenticate(route: Route?, response: Response): Request? {
```
hoangnguyen92dn, line 30: Just curious about this case, 2 requests return `401` at the same time and then we execute the mechanism to refresh the token but the requests for refreshing token executing in asynchronous, so 2 refresh tokens returned would be different 🤔

data/src/main/java/co/nimblehq/data/api/interceptor/AppRequestAuthenticator.kt
```diff
@@ -0,0 +1,83 @@
...
+                    .firstOrError()
+                    .blockingGet()
+                updateToken(refreshTokenResponse)
+
+                // Apply the authorization header with the new user token
+                val builder = response.request.newBuilder()
+                builder.removeHeader(AUTHORIZATION_HEADER)
+                val tokenType = securedStorage.userTokenType
+                val token = securedStorage.userAccessToken
+                builder.addHeader(AUTHORIZATION_HEADER, "$tokenType $token")
```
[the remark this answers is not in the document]
minhnimble: so cool to know, let me update 👍

data/src/main/java/co/nimblehq/data/api/interceptor/AppRequestAuthenticator.kt
minhnimble: hmm, let me think on how to handle the concurrency case 🤔

data/src/main/java/co/nimblehq/data/api/interceptor/AppRequestAuthenticator.kt
```diff
@@ -0,0 +1,83 @@
...
+        )
+        return try {
+            // Add max retry time as 3 here, to prevent the case it creates an infinite loop if backend keeps returning 401
+            if (refreshTokenRetryTime <= MAX_REFRESH_TOKEN_RETRY_TIME) {
+                refreshTokenRetryTime += 1
+
+                // Get new token from backend
+                val refreshTokenResponse = authService.refreshToken(refreshTokenRequest)
+                    .subscribeOn(RxScheduler.IoThread.get())
+                    .observeOn(RxScheduler.IoThread.get())
```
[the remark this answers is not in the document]
minhnimble: > I believe we could get rid of subscribeOn and observeOn

hmm I think this is recommended by @sleepylee. Can you please check with him on this? 🙏

> And why don't we update authService.refreshToken to Single rather than Flowable and then we could remove firstOrError, what do you think?

I think `Single` is okay for me too, I use `Flowable` because from the beginning I took references from The1 Android, but here since now I notice that our services will emit once per request and on demand data (most case of API) then they should be treated as a cold observable and `Single` is suited. We should use `Flowable` when the services are emitting items at a faster rate than the observer (app client) to benefit from the back pressure mechanism (hot observable)
If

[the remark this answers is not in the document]
hoangnguyen92dn: > hmm I think this is recommended by @sleepylee. Can you please check with him on this matter? 🙏

Aw yeah. I got Trung's idea here is the `refreshToken` will be executed in the background rather than the main thread.

Regarding to use `Single`, it depends on how you handle `Single` or `Flowable` in case sending a request to API error. Imo, every API requests would be `Single` so we don't need to call `firstOrError` :D You could keep the current implementation 😅

hoangnguyen92dn:
@minhnimble It will be better if you provide some UnitTest for the Authenticator 🚀

data/src/main/java/co/nimblehq/data/api/interceptor/AppRequestAuthenticator.kt
minhnimble: Handled the concurrent case, making sure the 2 unauthorized request have to wait for the refresh token complete before processing further, avoiding calling 2 refresh tokens 🙏

minhnimble:
> @minhnimble It will be better if you provide some UnitTest for the Authenticator 🚀

Will add another PR for unit test 🙏

data/src/main/java/co/nimblehq/data/api/authenticator/AppRequestAuthenticator.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    private var refreshTokenRetryTime = 1
+    private var isRefreshingToken = false
+
+    /**
+     * Authenticator for when the authToken need to be refresh and updated
+     * every time we get a 401 error code
+     */
+    @Throws(IOException::class)
+    @Synchronized
```
hoangnguyen92dn, line 32: `@Synchronized` -> It will block the `authenticate` method until everything in scope is done. Our purpose is just to prevent 2 API requests at the same time, why don't we just `synchronize` the block of API request?

data/src/main/java/co/nimblehq/data/api/authenticator/AppRequestAuthenticator.kt
```diff
@@ -0,0 +1,110 @@
...
+    override fun authenticate(route: Route?, response: Response): Request? {
+        return try {
+            // Add a wait check for the refresh token response while it is being loaded
+            while (isRefreshingToken) {
+                Thread.sleep(WAIT_CHECK_REFRESH_TOKEN_THRESHOLD_IN_MILLIS)
+            }
```
hoangnguyen92dn, line 38: By using synchronize, why do we still need to use this? 🤔

minhnimble: oh because even thought I use `synchronize` tag, it still triggers the requests twice. I forget to remove it, let me remove 🙏

data/src/main/java/co/nimblehq/data/api/authenticator/AppRequestAuthenticator.kt
minhnimble: hmm this is there because I am testing different ways to handle asynchronous call to `authenticate` but forget to remove, let me remove it 🙏

> why don't we just synchronize the block of API request?

Can you expand the idea a bit more, not sure how to just `synchronize` the API request block only? 🤔

hoangnguyen92dn: Sorry for the unclear explanation about my idea. Let me clarify again, we are concerned about when a request executing for refreshing a new token then another one coming and requesting a new token but the first one is not done yet. The issue here is when the first request received the new token success, but the API endpoint receives the second request for refreshing token so the first one would be invalid. Ideally, we just need to prevent sending refresh token requests and update new tokens on the app concurrency, and the rest of the mechanism in `authenticator` could be outside of the block. (Refer: https://www.geeksforgeeks.org/method-block-synchronization-java/)

minhnimble: thanks for sharing, let me investigate and apply this 🙏

minhnimble:
Hi @hoangnguyen92dn, I just complete update the branch with local `synchronize` only for the refresh token API call and the check whether we should refresh token again or not. Please help me take a look again 🙏

hoangnguyen92dn (approved):
LGTM

manh-t (approved):

